PR #3, opened 2016-10-07
Try to get the Libraries Factory to run with several instances

dankwart-de:


dankwart-de:
Whoohoo, don't know why and how this happend but it seems to be develop??

PR #5, opened 2016-10-10
Develop

vinjana:
A convertConfigurationValue(., ., Boolean) was created to ease testing without Roddy singleton object.


RoddyCore/src/de/dkfz/roddy/config/converters/BashConverter.groovy
```diff
@@ -198,9 +199,12 @@ class BashConverter extends ConfigurationConverter {
         return listOfSortedValues
     }
 
-    @Override
+    private boolean isQuoted(String string) {
+        (string.startsWith("'") && string.endsWith("'")) || (string.startsWith('"') && string.endsWith('"'))
+    }
+
     @CompileStatic
```
dankwart-de: CompileStatic can be removed. It is set for the whole class.

PR #9, opened 2017-01-09
Exec cache refactor

vinjana:
If outputAnalysisBaseDirectory (via --useiodir) is moved including the execCacheFile.txt, the path read from the execCacheFile.txt and the new --useiodir option caused an undescriptive stack trace. Now the error message (shown with the stack trace) includes more information.

Additional changes:
* more $dataSet variable checking (not just $pid)
* some code introduced during SSH debugging but still useful
* fix in compilation script (to allow working with symlinked source paths)


RoddyCore/test/de/dkfz/roddy/tools/RoddyIOHelperMethodsTest.groovy
```diff
@@ -179,4 +184,30 @@ public class RoddyIOHelperMethodsTest {
...
+    @Test
+    public void testSplitPathname() throws Exception {
+        assert RoddyIOHelperMethods.splitPathname("") == new ArrayList()
+        assert RoddyIOHelperMethods.splitPathname("/") == new ArrayList()
+        assert RoddyIOHelperMethods.splitPathname("/a/b") == ["a", "b"]
+        assert RoddyIOHelperMethods.splitPathname("//a/b/") == ["a", "b"]
+    }
+
+    @Test
+    public void testMatchComponentInPath() throws Exception {
```
dankwart-de: Would findComponentIndex be better than matchComponent? Actually you retrieve an index.

PR #10, opened 2017-01-09
Introduce a new feature for job output parameters

dankwart-de:
This commit changes the behaviour of the boolean "check" for job output file
checks. Instead of a simple boolean, we now will have a more
sophisticated conditional class. This can either be a pure boolean OR a
conditional.

Syntax is check="conditional:VARNAME"
or
check="true|false"

Errors in the definition will all fall back to true!

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
+++ b/RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
@@ -11,7 +11,9 @@ import de.dkfz.roddy.knowledge.files.BaseFile;
 import de.dkfz.roddy.knowledge.files.FileGroup;
 import de.dkfz.roddy.tools.BufferValue;
 import de.dkfz.roddy.tools.RoddyConversionHelperMethods;
+import de.dkfz.roddy.tools.RoddyIOHelperMethods;
 import de.dkfz.roddy.tools.TimeUnit;
+import examples.Exec;
 
 import java.io.File;
 import java.lang.reflect.Method;
@@ -217,6 +219,39 @@ public class ToolEntry implements RecursiveOverridableMapContainer.Identifiable 
         }
     }
 
+    public static class ToolFileParameterCondition {
+
+        private String condition;
+
+        private Boolean pureBoolean;
+
+        public ToolFileParameterCondition(boolean pureBoolean) {
+            this.pureBoolean = Boolean.valueOf(pureBoolean);
+        }
+
+        public ToolFileParameterCondition(String condition) {
+            if(condition.startsWith("conditional:"))
+                this.condition = condition.substring(12);
+            else {
+                this.pureBoolean = RoddyConversionHelperMethods.toBoolean(condition, true);
+            }
+        }
+
+        public boolean isPureBoolean() {
+            return pureBoolean != null;
+        }
+
+        public String getCondition() {
+            return condition;
+        }
+
+        public boolean evaluate(ExecutionContext context) {
+            if (pureBoolean != null)
+                return pureBoolean;
+            return context.getConfiguration().getConfigurationValues().getBoolean(condition, true);
+        }
+    }
+
     /**
      * Parameters for generic tools (tools which are not programmatically set!).
      * Parameters can be for in and for output.
@@ -227,15 +262,15 @@ public class ToolEntry implements RecursiveOverridableMapContainer.Identifiable 
         public final List<ToolConstraint> constraints;
         public final String scriptParameterName;
         public final String filenamePatternSelectionTag;
-        public final boolean checkFile;
+        public final ToolFileParameterCondition checkFile;
         public final String parentVariable;
         private List<ToolFileParameter> childFiles;
 
-        public ToolFileParameter(Class<BaseFile> fileClass, List<ToolConstraint> constraints, String scriptParameterName, boolean checkFile) {
+        public ToolFileParameter(Class<BaseFile> fileClass, List<ToolConstraint> constraints, String scriptParameterName, ToolFileParameterCondition checkFile) {
             this(fileClass, constraints, scriptParameterName, checkFile, FilenamePattern.DEFAULT_SELECTION_TAG, null, null);
         }
 
-        public ToolFileParameter(Class<BaseFile> fileClass, List<ToolConstraint> constraints, String scriptParameterName, boolean checkFile, String filenamePatternSelectionTag, List<ToolFileParameter> childFiles, String parentVariable) {
+        public ToolFileParameter(Class<BaseFile> fileClass, List<ToolConstraint> constraints, String scriptParameterName, ToolFileParameterCondition checkFile, String filenamePatternSelectionTag, List<ToolFileParameter> childFiles, String parentVariable) {
             super(scriptParameterName);
             this.fileClass = fileClass;
             this.constraints = constraints != null ? constraints : new LinkedList<>();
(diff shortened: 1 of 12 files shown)


RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
```diff
@@ -217,6 +219,39 @@ public ToolStringParameter clone() {
...
+        private String condition;
+
+        private Boolean pureBoolean;
+
+        public ToolFileParameterCondition(boolean pureBoolean) {
+            this.pureBoolean = Boolean.valueOf(pureBoolean);
+        }
+
+        public ToolFileParameterCondition(String condition) {
+            if(condition.startsWith("conditional:"))
```
vinjana: This is not the established type syntax, which is rather value:type or even name:value:type for the --cvalues CLI parameter. If at all, maybe better make the Boolean Type a subclass with fixed set of values ("true", "false").

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
```diff
@@ -217,6 +219,39 @@ public ToolStringParameter clone() {
...
+                this.condition = condition.substring(12);
+            else {
+                this.pureBoolean = RoddyConversionHelperMethods.toBoolean(condition, true);
+            }
+        }
+
+        public boolean isPureBoolean() {
```
vinjana: Why not just isBoolean? In what sense is this "pure"?

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
```diff
@@ -217,6 +219,39 @@ public ToolStringParameter clone() {
         }
     }
 
+    public static class ToolFileParameterCondition {
```
vinjana: Is this really a "condition"? Isn't this rather a "factor" (R-speak) upon which to condition?

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
```diff
@@ -227,15 +262,15 @@ public ToolStringParameter clone() {
...
         public final String parentVariable;
         private List<ToolFileParameter> childFiles;
 
-        public ToolFileParameter(Class<BaseFile> fileClass, List<ToolConstraint> constraints, String scriptParameterName, boolean checkFile) {
+        public ToolFileParameter(Class<BaseFile> fileClass, List<ToolConstraint> constraints, String scriptParameterName, ToolFileParameterCondition checkFile) {
             this(fileClass, constraints, scriptParameterName, checkFile, FilenamePattern.DEFAULT_SELECTION_TAG, null, null);
         }
 
-        public ToolFileParameter(Class<BaseFile> fileClass, List<ToolConstraint> constraints, String scriptParameterName, boolean checkFile, String filenamePatternSelectionTag, List<ToolFileParameter> childFiles, String parentVariable) {
+        public ToolFileParameter(Class<BaseFile> fileClass, List<ToolConstraint> constraints, String scriptParameterName, ToolFileParameterCondition checkFile, String filenamePatternSelectionTag, List<ToolFileParameter> childFiles, String parentVariable) {
```
vinjana: There is no checks, that checkFile is indeed a "boolean" instance of ToolFileParameterCondition.

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
```diff
@@ -217,6 +219,39 @@ public ToolStringParameter clone() {
...
+
+        public ToolFileParameterCondition(boolean pureBoolean) {
+            this.pureBoolean = Boolean.valueOf(pureBoolean);
+        }
+
+        public ToolFileParameterCondition(String condition) {
+            if(condition.startsWith("conditional:"))
+                this.condition = condition.substring(12);
+            else {
+                this.pureBoolean = RoddyConversionHelperMethods.toBoolean(condition, true);
```
vinjana: I think, the default "true" is not backwards compatible. Usually, empty values with Boolean were either false, or the default value obtained with toBoolean, which depended on the call site. With this, all unspecified values will be true. What we had as "Boolean" before was actually a tri-state: true, false, $defaultValue

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
dankwart-de: Groovy makes "everything" to a boolean. I had this somewhere in the back of my mind when I named it. pure means, this really is a boolean and nothing else.

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
dankwart-de: Actually this will not be used at all. But: The default value for check IS true. So I think in this case it is ok. When the value comes in via a configuration value, the normal checks are done upon config read-in.

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
dankwart-de: This is something different, you cannot pass it via cli! And it does not set a type. It just names the variable, which is used to define value of check.

<output ... check="condition:variablename" />

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
dankwart-de: Maybe the name is wrong.  It is the condition for check. If you do not want to set true/false, you can set a conditional variable instead.

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
dankwart-de: It does not necessarily need to be pure boolean. It can also hold a condition and will default to check when something fails.

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
vinjana: Maybe better "variable"? To me at least that sounds more intuitive.

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
vinjana: o.k.

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
vinjana: o.k.

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
vinjana: Alas, leave it if you wish.

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
vinjana: o.k.

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
dankwart-de: Done

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
dankwart-de: Done

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
dankwart-de: Done

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
dankwart-de: Renamed, done.

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
dankwart-de: Done.

PR #13, opened 2017-01-17
Line parse and split

dankwart-de:


RoddyCore/src/de/dkfz/roddy/tools/ComplexLine.groovy
```diff
@@ -0,0 +1,196 @@
...
+            } else if (fitsToOpeningCharacter(character, openedStarts.peek())) {
+                openedStarts.pop()
+                if (lineNodes.size() > 0)
+                    lineNodes.pop()
+                if (lineNodes.size() > 0)
+                    current = lineNodes.peek()
+                current.blocks << new TextNode(character as String);
+            } else {
+                ((TextNode) current.blocks[-1]).text << character;
+                openedStarts.push(character)
```
vinjana: There is no check here that this character is indeed an opening character. It could also be closing, namely if it is neither a normal character nor the closing character matching to the top of the openedStarts stack.

What would happen with the input 'blablabla)'?  (oops forgot the opening paren!) Should there not be at least a warning in this situation?

RoddyCore/test/de/dkfz/roddy/tools/ComplexLineTest.groovy
```diff
@@ -0,0 +1,123 @@
...
+    ]
+
+    public static final String codeLineExample = "def multiline='a string', b={something[in a braquet]}, method(a(b()))"
+    public static final String[] splittedCodeLineExample = [
+            "def multiline='a string'",
+            " b={something[in a braquet]}",
+            " method(a(b()))"
+    ]
+
+    public static final String faultyLineExample = "a(b(c())"
```
vinjana: I would like to see more aberrant inputs resulting in proper error messages or at least warnings (What is wrong? Which character in which string?). 

forgotten closing paren: "(blabla"
confused/non-matching paren types: "(a\\"b)\\""
forgotten opening paren: "balbla}"
non-matching paren types: "\\"blabla\\'"
singleton opening paren: "{"
singleton closing paren: ")"
etc.

Most of these inputs are wrong in 99% of the cases (even though they may not have to be fatal in the special case). But then, why should "(blabla" raise a parse error, but not "blabla)"? So if this is going to be a non-strict parser, what would be the conditions under which it should fail fatally and when to continue silently or only with a warning?

RoddyCore/test/de/dkfz/roddy/tools/ComplexLineTest.groovy
```diff
@@ -0,0 +1,123 @@
...
+    @Test
+    public void testParseCodeLine() {
+        ComplexLine.ComplexNode parsed = ComplexLine.parseLine(codeLineExample) as ComplexLine.ComplexNode;
+        assert parsed.blocks.size() == 7
+        assert parsed.blocks[0] instanceof ComplexLine.TextNode && parsed.blocks[0].reassemble() == "def multiline='"
+        assert ((ComplexLine.ComplexNode)parsed.blocks[1]).blocks[0].reassemble() == "a string"
+        assert parsed.blocks[2] instanceof ComplexLine.TextNode && parsed.blocks[2].reassemble() == "', b={"
+    }
+
+    @Test(expected = IOException)
```
vinjana: Can you make tests that also check the exception message? Look an @Rule.
Also: More faulty examples (see above)!

RoddyCore/src/de/dkfz/roddy/tools/ComplexLine.groovy
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        @Override
+        String reassemble() {
+            return text.toString()
+        }
+
+        @Override
+        String[] splitBy(String s) {
+            return text.toString().split(s)
```
vinjana: Maybe this would be a good place to handle special situations, such as ",blabla" or "blabla," when splitting on ','. Maybe you wouldn't need the extra + "" below then.

RoddyCore/src/de/dkfz/roddy/tools/ComplexLine.groovy
```diff
@@ -0,0 +1,196 @@
...
+            }
+            return _content.toString();
+        }
+
+        public String[] splitBy(String s) {
+            // Keep a list of splitted, prepared blocks. The list will be used to join results.
+            List<StringBuilder> splittedBlocks = []
+            for (int i = 0; i < blocks.size(); i++) {
+                List<String> splitted = null
+                if (blocks[i] instanceof TextNode) {
```
vinjana: Hm, would it be more appropriate to use a intuitively named, polymorphic (dynamic) method on TextNode and ComplexNode, that does the right thing, instead of the "instanceof"?

RoddyCore/src/de/dkfz/roddy/tools/ComplexLine.groovy
```diff
@@ -0,0 +1,145 @@
...
+        possibleContainersByOpeningCharacter['\'' as char] = '\'' as Character
+        possibleContainersByOpeningCharacter['\"' as char] = '\"' as Character
+        possibleContainersByOpeningCharacter['{' as char] = '}' as Character
+        possibleContainersByOpeningCharacter['[' as char] = ']' as Character
+        possibleContainersByOpeningCharacter['(' as char] = ')' as Character
+
+        possibleContainersByOpeningCharacter.each { Character key, Character value -> possibleContainersByClosingCharacter[value] = key }
+    }
+
+    String line;
```
dankwart-de: Done

RoddyCore/src/de/dkfz/roddy/tools/ComplexLine.groovy
```diff
@@ -0,0 +1,196 @@
...
+    LineNode parsedLine;
+
+    /**
+     * Parsed line represents a transformed / parsed line. Currently, single and double ticks will be used
+     * to identify children.
+     *
+     * E.g. --abc --def="ab,def;efk=test;anotherValue='equals'" --jki --a="something=def"
```
dankwart-de: Done

RoddyCore/src/de/dkfz/roddy/tools/ComplexLine.groovy
dankwart-de: Done, but not necessarily better.

RoddyCore/src/de/dkfz/roddy/tools/ComplexLine.groovy
```diff
@@ -0,0 +1,196 @@
...
+
+        public String[] splitBy(String s) {
+            // Keep a list of splitted, prepared blocks. The list will be used to join results.
+            List<StringBuilder> splittedBlocks = []
+            for (int i = 0; i < blocks.size(); i++) {
+                List<String> splitted = null
+                if (blocks[i] instanceof TextNode) {
+                    splitted = blocks[i].splitBy(s) as List<String>
+
+                    // If our string ends with the splitter, the splitted entry would be lost. Append an empty entry to preserve it.
```
dankwart-de: That's not a problem. ",,,".splitBy(",") will be [ "", "", "" ], that is also, why I put an extra "" at the end in some cases.

RoddyCore/src/de/dkfz/roddy/tools/ComplexLine.groovy
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    public static fitsToOpeningCharacter(Character c, Character o) {
+        return possibleContainersByClosingCharacter[c] == o
+    }
+
+    public static LineNode parseLine(String line) {
+
+        Stack<Character> openedStarts = new Stack<>();
+        openedStarts.push('ö' as char)
```
dankwart-de: Just needed something :p

RoddyCore/src/de/dkfz/roddy/tools/ComplexLine.groovy
```diff
@@ -0,0 +1,196 @@
...
+        Stack<ComplexNode> lineNodes = new Stack<>()
+        lineNodes.push(current)
+
+        for (int i = 0; i < line.length(); i++) {
+
+            char character = line.getChars()[i];
+
+            if (!isOpeningOrClosingCharacter(character)) {
+                ((TextNode) current.blocks[-1]).text << character;
+                continue;
```
dankwart-de: Done

RoddyCore/src/de/dkfz/roddy/tools/ComplexLine.groovy
```diff
@@ -0,0 +1,196 @@
...
+            } else if (fitsToOpeningCharacter(character, openedStarts.peek())) {
+                openedStarts.pop()
+                if (lineNodes.size() > 0)
+                    lineNodes.pop()
+                if (lineNodes.size() > 0)
```
dankwart-de: No! This is not the same! First I check if lineNode is > 0, then I pop from it and check again!

RoddyCore/src/de/dkfz/roddy/tools/ComplexLine.groovy
dankwart-de: Well it is either NOT an open/close OR a close OR finally an open.

RoddyCore/src/de/dkfz/roddy/tools/ComplexLine.groovy
```diff
@@ -0,0 +1,196 @@
...
+            } else {
+                ((TextNode) current.blocks[-1]).text << character;
+                openedStarts.push(character)
+                current.blocks << new ComplexNode()
+//                lineNodes.peek().children << current
+                current = current.blocks[-1] as ComplexNode
+                lineNodes.push(current);
+            }
+        }
+        if (lineNodes.size() > 1) throw new IOException("The line $line is malformed. There is an enclosing literal missing.")
```
dankwart-de: Yep

RoddyCore/test/de/dkfz/roddy/tools/ComplexLineTest.groovy
dankwart-de: Good question, I am tempted to let it be non-strict.

RoddyCore/src/de/dkfz/roddy/tools/ComplexLine.groovy
dankwart-de: Hm. Don't know, what you mean. I'd leave it for now?

RoddyCore/test/de/dkfz/roddy/tools/ComplexLineTest.groovy
dankwart-de: I'd leave it for now ok? I'll put some comments into the test class and see, how everything works out for the current cases.

RoddyCore/test/de/dkfz/roddy/tools/ComplexLineTest.groovy
dankwart-de: Done

PR #16, opened 2017-02-01
Additional check methods from QCPipeline put into ExecutionContext

dankwart-de:


RoddyCore/test/de/dkfz/roddy/core/ExecutionContextTest.groovy
```diff
@@ -0,0 +1,62 @@
...
+import de.dkfz.roddy.execution.io.fs.FileSystemAccessProvider
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+/**
+ * Created by heinold on 01.02.17.
+ */
+public class ExecutionContextTest {
```
vinjana: Do I see this right, that these methods are only part of the ExecutionContext, because they need it for error reporting? If so, would it not be more appropriate to just put it at the right place directly, and fix their interface, as soon as we have the error reporting refactored?

dankwart-de: No, I put them there, because the checkWorkflowExecutability method is also here. These methods are directly connected to the mentioned method.

PR #20, opened 2017-03-02
Bugfix: LocalExecutionService occasionally could not handle process I…

vinjana:
bugfix: LocalExecutionService

Additionally added slf4j updates to work with new SSHJ version.


vinjana:
These changes will be directly transferred into the new BatchEuphoria library.

PR #21, opened 2017-03-08
Readd FileGroup with fixed children support

dankwart-de:
This was removed as it behaves mostly like Tuples
For some older workflows we still need it. Don't want to break
compatibility here.


RoddyCore/src/de/dkfz/roddy/config/ToolFileParameter.java
```diff
@@ -50,6 +50,34 @@ public ToolFileParameter clone() {
         return new ToolFileParameter(fileClass, _con, scriptParameterName, checkFile, filenamePatternSelectionTag, childFiles, parentVariable);
     }
 
+    @Override
+    public boolean equals(Object o) {
```
vinjana: dito (hashCode)

RoddyCore/src/de/dkfz/roddy/knowledge/methods/GenericMethod.groovy
```diff
@@ -78,6 +78,15 @@ class GenericMethod {
     }
 
     /**
+     * If you need a file group output and the files are already set in the cfg, then this is the right method to call.
```
vinjana: What has this method to do with the config? The configs are not part of the interface, so this is not obvious.

RoddyCore/test/de/dkfz/roddy/config/ToolFileParameterTest.groovy
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Created by heinold on 08.03.17.
+ */
+@CompileStatic
+public class ToolFileParameterTest {
+    @Test
+    public void testCloneAndEquals() throws Exception {
+        def src = new ToolFileParameter(BaseFile, [], "ABC", new ToolFileParameterCheckCondition(true), "DEF", null, "abc")
+        def clone = src.clone()
+        assert clone == src
```
vinjana: ??? Why are you testing `==` and not `.equals()`? Thus src.clone() could/should return `this`. Also, the issue with the `scriptparameter` can not be found. There is also no negative controls here! You'd need to test with .equals() and have tests in which each one of the values is different.

RoddyCore/src/de/dkfz/roddy/config/ConfigurationFactory.groovy
```diff
@@ -924,22 +924,39 @@ public class ConfigurationFactory {
...
         ToolFileGroupParameter.IndexOptions indexOptions = Enum.valueOf(ToolFileGroupParameter.IndexOptions.class, extractAttributeText(groupNode, "indices", ToolFileGroupParameter.IndexOptions.numeric.name()))
 
         String fileclass = extractAttributeText(groupNode, "fileclass", null);
-        Class genericFileClass = LibrariesFactory.getInstance().loadRealOrSyntheticClass(fileclass, BaseFile.class)
+        int childCount = groupNode.children().size();
         String pName = groupNode.@scriptparameter.text();
 
+        if (fileclass) {
+            Class genericFileClass = LibrariesFactory.getInstance().loadRealOrSyntheticClass(fileclass, BaseFile.class)
+            if (!pName || !fileclass)
```
dankwart-de: Done

RoddyCore/src/de/dkfz/roddy/config/ConfigurationFactory.groovy
```diff
@@ -924,22 +924,39 @@ public class ConfigurationFactory {
...
-        if(!pName || !fileclass)
-            throw new RuntimeException("You have to set both the parametername and the fileclass attribute for filegroup i/o parameter in ${toolID}")
-
-        ToolFileGroupParameter tpg = new ToolFileGroupParameter(filegroupClass, genericFileClass, pName, passas, indexOptions)
-        return tpg
+    @Deprecated
+    private static ToolFileGroupParameter parseChildFilesForFileGroup(NodeChild groupNode, PassOptions passas, String toolID, String pName, Class filegroupClass, ToolFileGroupParameter.IndexOptions indexOptions) {
+        int childCount = groupNode.children().size();
+        List<ToolFileParameter> children = new LinkedList<ToolFileParameter>();
+        if (childCount == 0 && passas != PassOptions.array)
```
dankwart-de: This handles how the files are passed to the jobs. We can talk about this again.

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
```diff
@@ -44,6 +44,17 @@ public void apply(BaseFile p) {
         public ToolConstraint clone() {
             return new ToolConstraint(onFailMethod, checkMethod);
         }
+
+        @Override
+        public boolean equals(Object o) {
```
dankwart-de: Didn't know about that rule :D Will do it.

RoddyCore/src/de/dkfz/roddy/config/ToolFileGroupParameter.groovy
```diff
@@ -40,11 +53,38 @@ class ToolFileGroupParameter extends ToolEntry.ToolParameter<ToolFileGroupParame
         this.indexOptions = indexOptions
     }
 
-
+    /**
+     * Copy constructor
```
dankwart-de: It is used as such

RoddyCore/src/de/dkfz/roddy/config/ToolFileGroupParameter.groovy
```diff
@@ -40,11 +53,38 @@ class ToolFileGroupParameter extends ToolEntry.ToolParameter<ToolFileGroupParame
...
+    }
 
     @Override
     ToolFileGroupParameter clone() {
-        return new ToolFileGroupParameter(groupClass, genericFileClass, scriptParameterName, passOptions)
+        return new ToolFileGroupParameter(scriptParameterName, groupClass, genericFileClass,  passOptions, indexOptions, files?.collect { ToolFileParameter tfp -> tfp.clone() })
+    }
+
+    @Override
+    boolean equals(o) {
```
dankwart-de: Don't know... in equals super.equals should be checked, right? I did it for this.

RoddyCore/src/de/dkfz/roddy/knowledge/methods/GenericMethod.groovy
dankwart-de: Well where should I put them? These are convenience methods. And filegroups need to be treated different than regular calls.

RoddyCore/test/de/dkfz/roddy/config/ToolFileParameterTest.groovy
dankwart-de: Groovy == == .equals right?

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
vinjana: Well, it's a rule, not more. I think in a hash, the hashCode is used for the bucket, but within the bucket the equals() method is used to determine equality. Therefore, having different hashCodes but being equal(s) can lead to problems. hashCode() is kind of part of equality in the hash context.

But as I said, if we know, that we won't use the hashCode, we are safe.

RoddyCore/src/de/dkfz/roddy/config/ToolFileGroupParameter.groovy
vinjana: But super.equals(scriptparameter) is not called in this method.

vinjana: Note that ToolParameter does not even have an .equals() method defined by you. Therefore it will use the reference and not the scriptParameterName to determine equality.

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
```diff
@@ -44,6 +44,17 @@ public void apply(BaseFile p) {
...
+            if (o == null || getClass() != o.getClass()) return false;
+
+            ToolConstraint that = (ToolConstraint) o;
+
+            if (onFailMethod != null ? !onFailMethod.equals(that.onFailMethod) : that.onFailMethod != null) return false;
+            return checkMethod != null ? checkMethod.equals(that.checkMethod) : that.checkMethod == null;
+        }
     }
 
     public static abstract class ToolParameter<T extends ToolParameter> {
```
vinjana: There is no .equals() method defined by you (but equals() is used in subclass)

RoddyCore/src/de/dkfz/roddy/config/ToolFileGroupParameter.groovy
vinjana: In Groovy the method declared in a subclass with the same signature as the same-named method in the superclass overrides the superclass's method, right? So super.equals() is not implicitly called.

dankwart-de: Yes, it is not implicetly called. I think I will take a look on how Java works with equals to clarify things.

RoddyCore/test/de/dkfz/roddy/config/ToolFileParameterTest.groovy
vinjana: AFAIKT in Java .equals is **per default** set to ==. But indeed you are right in Groovy the situation is different. But apparently it is not as simple as you said (and is written here by mrhaki in 2009: http://mrhaki.blogspot.de/2009/09/groovy-goodness-check-for-object.html). 

See for instance this somewhat newer post: http://stackoverflow.com/questions/13687154/groovy-operator.

I guess we should have a look at the source code for equals and ==. :-|

dankwart-de: Ooook... Good I hope things will not get more complicated now. I'll take a look at the post later.

RoddyCore/src/de/dkfz/roddy/config/ToolEntry.java
```diff
@@ -49,12 +49,20 @@ public ToolConstraint clone() {
         public boolean equals(Object o) {
             if (this == o) return true;
```
vinjana: Use explicit reference equality (also below; just in case we transform this later into Groovy, where == would result in infinite recursion.

PR #23, opened 2017-03-13
Bugfix null pointer on plugin version collision

vinjana:
Is an error actually the right way to proceed, or should we warn and rename the file?


vinjana:
As discussed, @dankwart-de, this can be merged after update to current devel.

PR #27, opened 2017-03-16
Logger updates

dankwart-de:


RoddyCore/src/de/dkfz/roddy/client/RoddyStartupOptions.groovy
```diff
@@ -30,6 +30,8 @@ public enum RoddyStartupOptions {
     useconfig(true),
     usefeaturetoggleconfig(true),
     verbositylevel(true),
+    v(false), // Verbosity of 3
```
vinjana: document in usage info (help)

PR #28, opened 2017-03-21
Add initial checks, Fix cvalue dependency bug, Add --rv parameter

dankwart-de:
The method performInitialCheck() in Roddy.java is introduced. Currently
it checks for existens of jar/zip/unzip using which! Not ideal but a
start.

There was a bug for configuration values:
     * When you have several configuration objects, e.g.
     * A -> B
     *
     * A contains values a and b, where a == abc and b has a reference
     * to a b=${a}
     * B contains values a'  == def
     * If you query B for b, b holds a reference to configuration A
     * Therefore, b.toString() will result in b == abc and NOT in def as
     * expected.
     *
     * By elevating a copy of b to B, everything will be fixed, as B can
     * be the parent of b'
     *
     * This is true for ConfigurationValue, which rely on dependencies
     * between themselves.
     *
     * In RecursiveOverridableMapContainerForConfigurationValues, we
     * override the method and
     * copy the value and set the configuration reference to (in this
     * example) B

Add --rv for --useRoddyVersion, rework help message

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/config/RecursiveOverridableMapContainer.java
+++ b/RoddyCore/src/de/dkfz/roddy/config/RecursiveOverridableMapContainer.java
@@ -20,15 +20,14 @@ public class RecursiveOverridableMapContainer<K, V extends RecursiveOverridableM
 
 
 
-    public static interface Identifiable {
-        public String getID();
-
+    public interface Identifiable {
+        String getID();
     }
 
     /**
      * A list of values in this container's configuration
      */
-    private final Map<K, V> values = new LinkedHashMap<>();
+    protected final Map<K, V> values = new LinkedHashMap<>();
 
 
     private final P containerParent;
@@ -105,17 +104,53 @@ public class RecursiveOverridableMapContainer<K, V extends RecursiveOverridableM
         return getListOfAllValues().contains(id);
     }
 
+    /**
+     * When you have several configuration objects, e.g.
+     * A -> B
+     *
+     * A contains values a and b, where a == abc and b has a reference to a b=${a}
+     * B contains values a'  == def
+     * If you query B for b, b holds a reference to configuration A
+     * Therefore, b.toString() will result in b == abc and NOT in def as expected.
+     *
+     * By elevating a copy of b to B, everything will be fixed, as B can be the parent of b'
+     *
+     * This is true for ConfigurationValue, which rely on dependencies between themselves.
+     *
+     * In RecursiveOverridableMapContainerForConfigurationValues, we override the method and
+     * copy the value and set the configuration reference to (in this example) B
+     *
+     * @param src
+     * @return
+     */
+    protected V copyValueToThisContainer(V src) {
+        add(src);
+        return src;
+    }
+
+    /**
+     * Does not check, if the value exists!
+     * @param id
+     * @return
+     */
+    protected V _getValueUnchecked(String id) {
+        V cval = getAllValues().get(id);
+        if(!values.containsKey(id))
+            return copyValueToThisContainer(cval);
+
+        return cval;
+    }
+
     public V getValue(String id, V defaultValue) {
         if (!hasValue(id))
             return defaultValue;
-        return getAllValues().get(id);
+        return _getValueUnchecked(id);
     }
 
     public V getValue(String id) {
         if (!hasValue(id))
             throw new RuntimeException("Value " + id + " could not be found in containers with id " + id);
-
-        return getAllValues().get(id);
+        return _getValueUnchecked(id);
     }
 
     public void putAll(Map<K, V> values) {
@@ -134,6 +169,14 @@ public class RecursiveOverridableMapContainer<K, V extends RecursiveOverridableM
     public void addAll(List<V> all) {
         if(RoddyConversionHelperMethods.isNullOrEmpty(all)) return;
         for(V it : all) { add(it); }
+    }
+
+    public void append(V value) {
+        add(value);
+    }
+
+    public void leftShift(V value) {
+        add(value);
     }
 
     public int size() {
--- a/RoddyCore/src/de/dkfz/roddy/config/RecursiveOverridableMapContainer.java
+++ b/RoddyCore/src/de/dkfz/roddy/config/RecursiveOverridableMapContainer.java
@@ -18,8 +18,6 @@ import java.util.Map;
  */
 public class RecursiveOverridableMapContainer<K, V extends RecursiveOverridableMapContainer.Identifiable, P extends ContainerParent> {
 
-
-
     public interface Identifiable {
         String getID();
     }
@@ -28,7 +26,6 @@ public class RecursiveOverridableMapContainer<K, V extends RecursiveOverridableM
      * A list of values in this container's configuration
      */
     protected final Map<K, V> values = new LinkedHashMap<>();
-
 
     private final P containerParent;
 
@@ -123,8 +120,7 @@ public class RecursiveOverridableMapContainer<K, V extends RecursiveOverridableM
      * @param src
      * @return
      */
-    protected V copyValueToThisContainer(V src) {
-        add(src);
+    protected V temporarilyElevateValue(V src) {
         return src;
     }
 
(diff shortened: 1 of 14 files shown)


vinjana:
How did the reference to `a` get into `b` in the first place? Could this not be considered as a parsing problem?

vinjana:
So you mean you have two configuration objects `A` and `B`, where

`A = { a = "abc" }`
`B = { a = "def", b = "$A.a" }`

A references B: `A -> B`

Then B.b.toString() = "abc", but you expect that the reference `$A.a` actually should be `$B.a`, because value `a` can be satisfied with the value in `B`. Right?

What do you mean by "elevating a copy of b to B"?

dankwart-de:
You nearly got it right. 
A = { a = "abc", b="${a}" }
B = { a' = "def" }

so B.b.toString() = "def"

Elevating means, that B will get a copy of b, so it can properly resolve dependencies. In your syntax:
A = { a = "abc", b="${a}" }
B = { a' = "def", b'="${a}" }

vinjana:
I guess the behaviour that you try to achieve is:

- There is precendence between configuration objects. `A -> B` means that a variable `a` defined in `A` will override the variable `a` of the same name in `B`. `A -> B` defines a "variable stack".
- When a variable is not contained in `A`, the variable in `B` will be returned.
- When a variable `b` references a variable `a`, e.g. written as `b = "$a"` the request will also be satisfied by the first variable value in the variable stack `A -> B`.

In particular the last rule would suggest that one simply calls `A.a` to get the globally evaluated value for `a`, at what level in the stack it may be. Right?

dankwart-de:
I think we mixed up directions:
A is the base configuration. So initiallay a and b are defined in A.
B is the extended configuration (A -> B) and contains a'

B.b should then return A.b with the content from B.a'. However this did not happen, as b only knew about A and a. Elevating b to b' makes it possible for resolve all dependencies.

vinjana:
What would happen in the case `A -> B -> C` where
```
A = { a = "abc", b = "def" }
B = { a = "hij", b = "$a" }
C = { a = "klm" }
```
`C.b = "klm"` or `C.b = "hij"`?

RoddyCore/src/de/dkfz/roddy/config/RecursiveOverridableMapContainer.java
```diff
@@ -105,17 +104,53 @@ public boolean hasValue(String id) {
...
+
+    /**
+     * Does not check, if the value exists!
+     * @param id
+     * @return
+     */
+    protected V _getValueUnchecked(String id) {
+        V cval = getAllValues().get(id);
+        if(!values.containsKey(id))
+            return copyValueToThisContainer(cval);
```
vinjana: Hm, this adds a side-effect to a get-operation. I'm not happy with this. Given our discussion, would be querying the full tree be an alternative?

dankwart-de:
In that case C.b == klm. That is what is expected. I'll create a test and add that.

vinjana:
Do I see it correctly, this is like evaluating all variables in `A -> B -> C` (so C overrides B overrides A) as if querying the whole configuration tree again from C? So in the example above

```
A = { a = "abc", b = "def" }
B = { a = "hij", b = "$a" }
C = { a = "klm" }
```

What is actually supposed to happen is that when a variable `a` is queried, actually `a* := C.a` is requested, which will be recursively evaluated up to the point, where `a` can be matched. But when `a` itself refers to a variable, say `b`, also that variable value is obtained by evaluating from `C`, that is `b* := C.b`. Right?

RoddyCore/test/de/dkfz/roddy/config/ConfigurationValueTest.groovy
```diff
@@ -13,15 +13,26 @@ import org.junit.Test
  * Created by heinold on 14.07.16.
  */
 @CompileStatic
-class ConfigurationValueTest extends GroovyTestCase {
-
-    // Test for determineTypeOfValue is in RoddyConversionHelperMethodsTest.groovy
-    // The method is directly bound to several methods of the conversion class and has nearly no
-    // logic of its own. So we put it there.
+class ConfigurationValueTest {
```
vinjana: This configuration variable evaluation is extremely important (what happens if the evaluation is wrong and say 1000 WGS samples are processed the wrong way?!

I suggest we write many more tests -- and, actually, for the old version! After that we test that the behaviour of preserved for this new version. I also suggest we read in a CO configuration tree and few plugin configurations, and check before and after the change, to ensure that things go as they should. I feel extreme uncomfortable changing this code without tests.

RoddyCore/test/de/dkfz/roddy/config/ConfigurationValueTest.groovy
```diff
@@ -13,15 +13,137 @@ import org.junit.Test
...
 @CompileStatic
-class ConfigurationValueTest extends GroovyTestCase {
+class ConfigurationValueTest {
 
-    // Test for determineTypeOfValue is in RoddyConversionHelperMethodsTest.groovy
-    // The method is directly bound to several methods of the conversion class and has nearly no
-    // logic of its own. So we put it there.
+    /**
+     * Given are two configurations:
+     * C extends B, B extends A
```
vinjana: C?

RoddyCore/test/de/dkfz/roddy/config/ConfigurationValueTest.groovy
```diff
@@ -13,15 +13,137 @@ import org.junit.Test
...
+     * C1 = { c = 'C.c' }                           <br/>
+     *
+     * A1.b = 'A1.a'
+     * A2.b = 'A2.b'
+     *
+     * B1.a = 'A2.a'
+     * B1.b = 'A2.b'
+     * B1.d = 'A2.a'
+     *
+     * B2.a = 'B2.b + ${c}'
```
vinjana: Maybe add: `c` remains unevaluated, as there is no `c` defined in the `B1`, `A1`, or `A2`.

RoddyCore/test/de/dkfz/roddy/config/ConfigurationValueTest.groovy
```diff
@@ -13,15 +13,137 @@ import org.junit.Test
...
+     *
+     * C1.a = 'B2.b + C.c'
+     * C1.b = 'B2.b'
```
vinjana: This means, that the priority is `B2 > B1`, right?

BTW: Say, C1.b would evaluate to (=>) B1.b, then

 `C1.b => B1.b = "$a" => B2.a = "$b + $c"` => "$b + C.c"`

But the last `$b` again refers to C1.b, so there would be a loop!

We have no way to catch this, in the moment, right? I guess we need to make a check and write an error, if there is a cyclic definition.

RoddyCore/test/de/dkfz/roddy/config/ConfigurationValueTest.groovy
```diff
@@ -13,15 +13,137 @@ import org.junit.Test
...
     @Test
-    public void testNothing() {
-        // Asimple test so the class really fails.
-        assert false
+    void testCValueSubstitutionWithComplexHierarchy() {
+        Configuration A1 = new Configuration(null)
+        Configuration A2 = new Configuration(null)
+        Configuration B1 = new Configuration(null, [A1, A2])
+        Configuration B2 = new Configuration(null)
+        Configuration C = new Configuration(null, [B1, B2])
```
vinjana: Note that you provide `[B1, B2]`. Is that on purpose that `B2` has precedence over `B1`? Intuitively, I would have thought `B1 > B2`, because it is first in the list. In any case, it should be documented also with the Configuration constructor, such that cannot go wrong there.

dankwart-de: done

RoddyCore/test/de/dkfz/roddy/config/ConfigurationValueTest.groovy
dankwart-de: Nope, cyclic references are not cached yet.

RoddyCore/test/de/dkfz/roddy/config/ConfigurationValueTest.groovy
dankwart-de: done

RoddyCore/test/de/dkfz/roddy/config/ConfigurationValueTest.groovy
dankwart-de: done

PR #35, opened 2017-04-03
Fixed fillConfigurationVariables

dankwart-de:
The method stopped, when a value was not in the configuration values
list. Now it will blacklist it and fill it back in at the end of the
process.


RoddyCore/src/de/dkfz/roddy/client/RoddyStartupModes.groovy
```diff
@@ -111,8 +111,8 @@ enum RoddyStartupModes {
         System.out.println(String.format("  %s %s " + Constants.ENV_LINESEPARATOR + "\t%s" + Constants.ENV_LINESEPARATOR, option.toString(), parameters, formattedDescription.toString()))
     }
 
-    static void printCommands(List<Object[]> commands) {
-        commands.each { Object[] it ->
+    static void printCommands(List<List<Object>> commands) {
+        commands.each { List<Object> it ->
             printCommand(it[0] as RoddyStartupModes, it[1] as String, it[2..-1] as String[])
```
fkaercher: A list of objects could be anything. This object type casting is not safe.

RoddyCore/src/de/dkfz/roddy/config/FilenamePattern.java
```diff
@@ -228,9 +232,21 @@ String fillConfigurationVariables(String src, ExecutionContext context) {
             // Only change it, if the value is in the configuration.
             if(configurationValues.hasValue(command.rawName)) {
                 src = src.replace(command.fullString, configurationValues.get(command.rawName).toString());
+            } else {
+                String value = "###" + blacklistID + "###";
```
fkaercher: What's the purpose of the blacklist?

RoddyCore/src/de/dkfz/roddy/client/RoddyStartupModes.groovy
dankwart-de: Unfortunately, I could not come up with a type safe version without adding something like
class CommandPrintArgument {
RoddyStartupModes mode;
String options;
String[] additionalMessages;
}

Would be a lot more code. Suggestions? We could maybe convert it to List<String>. That is still not type safe but we convert every entry to string anyways.

RoddyCore/src/de/dkfz/roddy/config/FilenamePattern.java
dankwart-de: The purpose is to prevent the while loop to check for the same value again and again. If a variable is used but cannot be found in the config, it is blacklisted. In the next iteration it is not looked at.

PR #36, opened 2017-04-04
Clean up startup messages, correct cli toggle input

dankwart-de:
It is not only about colours. But the testrun/rerun where cleaned up.

Also toggle values are now parsed in a correct way. If a toggle value
does not exist, an error is shown and a list of available toggles is
printed.

If the ini file is not found, Roddy will not start!

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/Roddy.java
+++ b/RoddyCore/src/de/dkfz/roddy/Roddy.java
@@ -229,7 +229,7 @@ public class Roddy {
         LoggerWrapper.setup();
         logger.postAlwaysInfo("Roddy version " + Constants.APP_CURRENT_VERSION_STRING);
 
-        if(!performInitialCheck())
+        if (!performInitialCheck())
             exit(1);
 
         time("ftoggleini");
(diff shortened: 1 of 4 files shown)


RoddyCore/src/de/dkfz/roddy/Roddy.java
```diff
@@ -399,19 +399,39 @@ private static void initializeFeatureToggles() {
...
+            return true;
         }
+
+        // Stores if an error happened
+        boolean error = false;
+        if (commandLineCall.isOptionSet(opt)) {
+            List<String> listOfToggles = commandLineCall.getOptionList(opt);
+            for(String toggle : listOfToggles) {
+                try {
+                    if (AvailableFeatureToggles.valueOf(toggle) != null)
```
dankwart-de: Yes you're right, but that is exactly what is happening three lines down.
Actually the query throws an exception but I did not come up with a short way to make a safe version in Java.

However, if you look into the catch part, there is the error message and the toggle name.
Also, if the method returns false, a list of the available toggles will be printed.

PR #39, opened 2017-04-26
Cleaned load errors, Fix cyclic depencies

dankwart-de:
Cyclic dependencies are now detected. The mechanism is somewhat crude
but should work.

Cleaned up the error throwing mechanism for configuration loader errors.
SaxParseExceptions will be catched and converted to a
ProjectLoaderException thus resulting in less stack trace printouts.
Unknown errors will still be printed fully.


dankwart-de:
Does actually not work.

PR #47, opened 2017-05-03
Fix job abortion on workflow errors

dankwart-de:
The abort command will now only be called, when an exception occurs.

Another logger message will be shown, stating what you can do to enable
strict mode.

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/core/Analysis.java
+++ b/RoddyCore/src/de/dkfz/roddy/core/Analysis.java
@@ -389,8 +389,10 @@ public class Analysis {
                         context.execute();
                         finallyStartJobsOfContext(context);
                     }
-                } finally {
+                } catch(Exception ex) {
+                    // (Maybe) abort jobs in strict mode
                     abortStartedJobsOfContext(context);
+                } finally {
 
                     if (context.getExecutionContextLevel() == ExecutionContextLevel.QUERY_STATUS) { //Clean up
                         //Query file validity of all files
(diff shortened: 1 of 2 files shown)


RoddyCore/src/de/dkfz/roddy/core/Analysis.java
```diff
@@ -389,8 +389,10 @@ protected void executeRun(ExecutionContext context, boolean preventLoggingOnQuer
                         context.execute();
                         finallyStartJobsOfContext(context);
                     }
-                } finally {
+                } catch(Exception ex) {
```
vinjana: space before (  

(I'm only commenting on this because this is a pretty frequent change in the recent commits)

RoddyCore/src/de/dkfz/roddy/core/Analysis.java
```diff
@@ -389,8 +389,10 @@ protected void executeRun(ExecutionContext context, boolean preventLoggingOnQuer
...
+                    // (Maybe) abort jobs in strict mode
                     abortStartedJobsOfContext(context);
```
vinjana: If this is "(maybe)", why not call it `maybeAbortSTartedJobsOfContext(context)`?

RoddyCore/src/de/dkfz/roddy/core/Analysis.java
```diff
@@ -496,6 +498,9 @@ private void abortStartedJobsOfContext(ExecutionContext context) {
             } catch (Exception ex) {
                 logger.severe("Could not successfully abort jobs.", ex);
             }
+        } else {
+            logger.severe("An workflow error occurred. However, strict mode is disabled and/or RollbackOnWorkflowError is disabled and therefore, all submitted jobs will be left running." +
```
vinjana: *A* workflow error ...

I like the warning, if things continue. Really seems to be a 180 degree turn in strategy. I am pretty sure this will be an improvement :-)

PR #69, opened 2017-06-29
Project config free mode

dankwart-de:


RoddyCore/src/de/dkfz/roddy/Roddy.java
```diff
@@ -820,7 +820,9 @@ public static File getCompressedAnalysisToolsOverviewFile() {
     }
 
     public static List<File> getConfigurationDirectories() {
-        return loadFolderListFromConfiguration(RoddyStartupOptions.configurationDirectories, Constants.APP_PROPERTY_CONFIGURATION_DIRECTORIES);
+        List<File> list = loadFolderListFromConfiguration(RoddyStartupOptions.configurationDirectories, Constants.APP_PROPERTY_CONFIGURATION_DIRECTORIES);
+        list.add(getFolderForConfigurationFreeMode());
```
vinjana: Hm, does this ensure that the explicit configuration still works? Is there a chance that latter configs in the list override earlier?

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -34,6 +37,12 @@ class ProjectLoader {
      */
     private static Map<String, Analysis> _analysisCache = [:]
 
+    private static final int PSRC_UNSET = 0
```
vinjana: PSRC = Project Config Source? Maybe no abbrev.?
Also this is a typical task for an enum, isn't it?

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -311,10 +384,16 @@ class ProjectLoader {
     }
 
     String getFullAnalysisID(String projectID, String analysisID) {
+//        if (projectSource == PSRC_PLUGIN) {
```
vinjana: ?

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -169,6 +192,56 @@ class ProjectLoader {
...
+
+            projectID = "CFreeMode_" + Integer.toHexString(configurationFileName.hashCode())
+
+            File configurationFile = new File(Roddy.getFolderForConfigurationFreeMode(), configurationFileName)
+
+            logger.always("Will use automatically generated configuration file: ${configurationFileName}")
+
+            String configurationText = lines.join("\n")
+            if (configurationFile.exists()) {
+                if (RoddyIOHelperMethods.getMD5OfText(configurationText) != RoddyIOHelperMethods.getMD5OfFile(configurationFile)) {
```
vinjana: Hm, I think this should not happen silently. Do we want to have a copy of the old file, or is a "copy" of the config in the output paths sufficient?

And why do you compare the MD5s instead of the texts? For the md5s the texts has to be traversed fully, for the comparison not. And with a plain comparison the code would get simpler.

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -169,6 +192,56 @@ class ProjectLoader {
             throw new ProjectLoaderException("Could not extract analysis for ${configurationIdentifier}.");
     }
 
+    void checkAndPreparePremisesOrFail() {
```
vinjana: Pretty long method. Can it be broken down?

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
dankwart-de: At the moment I think that this check is invalid.:
- I did not implement any file locking mechanisms which would will prevent i/o hazards. Need to do that but I don't know yet if Java can do this across processes.
- It is not neccessary to write out a file with the same md5 sum.

Well comparing the md5sum is easy as it is only two commands to have two sums. Also the files will be very small, so I don't think it matters.

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
dankwart-de: I'll try that.

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
dankwart-de: Yes this might be an enum task. I tend to use this syntax for smaller "enums"

RoddyCore/src/de/dkfz/roddy/Roddy.java
dankwart-de: What do you mean exactly? Files are identified by an id and Roddy will tell you, when there's a duplicate.

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
dankwart-de: I can remove it.

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
dankwart-de: Switched it to an enum which is also more describing.

RoddyCore/src/de/dkfz/roddy/Roddy.java
dankwart-de: Acutally this is unlikely, the id is quite long and also uses a hash code.

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
dankwart-de: done. Extracted the code for configuration file creation

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
dankwart-de: Integrated file locking with channels

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
dankwart-de: done

PR #77, opened 2017-07-11
Changes according to PR #69

dankwart-de:
Cannot open #69  again.

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
+++ b/RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
@@ -218,7 +218,7 @@ class ProjectLoader {
             return
         }
 
-        throw new ProjectLoaderException("The project identifier is wrong it must either be a configuration or a plugin. Properly set projectConfigurationSource in the ProjectLoader.")
+        throw new ProjectLoaderException("The project identifier is wrong it must either be a configuration identifier or a plugin name. Properly set projectConfigurationSource in the ProjectLoader.")
     }
 
     private void createAndWriteAutoConfigurationFile() {
@@ -245,7 +245,7 @@ class ProjectLoader {
 
         // Create a JVM wide file lock to allow only one process which writes to the target directory.
         RandomAccessFile directoryLockFile = new RandomAccessFile(new File(Roddy.getFolderForConfigurationFreeMode(), ".configLock"), "rw")
-        FileLock flock = directoryLockFile.channel.lock()
+        FileLock jvmWideDirectoryLockFile = directoryLockFile.channel.lock()
 
         try {
             String configurationText = lines.join("\n")
@@ -259,7 +259,7 @@ class ProjectLoader {
             }
         } finally {
             // Unlock and close the file
-            flock.release()
+            jvmWideDirectoryLockFile.release()
             directoryLockFile.close()
         }
     }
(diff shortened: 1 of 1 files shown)


RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -193,40 +200,54 @@ class ProjectLoader {
...
                 throw new ProjectLoaderException("It is not possible to use the project configuration free mode without the --${RoddyStartupOptions.useiodir.name()} option.")
 
-            // Create a temporary project configuration in ~/.roddy/configurationFreeMode
-            String configurationFileName = "projects" + projectID + "_" + analysisID + "_" + Integer.toHexString(Roddy.commandLineCall.getOptionValue(RoddyStartupOptions.useiodir).hashCode()) + ".sh"
+            createAndWriteAutoConfigurationFile()
+
+            return
+        }
+
+        throw new ProjectLoaderException("The project identifier is wrong it must either be a configuration or a plugin. Properly set projectConfigurationSource in the ProjectLoader.")
```
vinjana: "... configuration identifier or plugin name ..." ? Would be slightly clearer, or not?

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -193,40 +200,54 @@ class ProjectLoader {
...
-            lines << "#analysis " + analysisImport
-            if (Roddy.isOptionSet(RoddyStartupOptions.baseconfig)) lines << "#imports " + Roddy.commandLineCall.getOptionValue(RoddyStartupOptions.baseconfig)
+        lines << "inputBaseDirectory=" + Roddy.customBaseInputDirectory
+        lines << "outputBaseDirectory=" + Roddy.customBaseOutputDirectory
+        lines << 'outputAnalysisBaseDirectory=${outputBaseDirectory}/${pid}'
 
-            lines << "inputBaseDirectory=" + Roddy.customBaseInputDirectory
-            lines << "outputBaseDirectory=" + Roddy.customBaseOutputDirectory
-            lines << 'outputAnalysisBaseDirectory=${outputBaseDirectory}/${pid}'
+        projectID = "CFreeMode_" + Integer.toHexString(configurationFileName.hashCode())
```
vinjana: Not your fault, but github's differ really sucks here :-(

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -236,10 +257,11 @@ class ProjectLoader {
             } else {
                 configurationFile << configurationText
             }
-            return
+        } finally {
+            // Unlock and close the file
+            flock.release()
```
vinjana: * Why do we need this?
* flock.release() cannot fail?
* Is this the flock syscall in Linux? We know it is not working well on NFS4 and in OTP we actually removed all flock-calls to be more portable and become able to use NFS4 and thus get able to use extended attributes and POSIX ACLs. Reconsider.

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
dankwart-de: Yes it is. I'll change it.

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
dankwart-de: One more thing which causes you scream and run in circles... :D

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
dankwart-de: Ok, I'll put in a more sophisticated comment. The idea behind channel locks is, that the lock is JVM wide and therefore shared across all JVM instances on a single machine. Not more not less. It will just prevent the current user from mutual file access / file overrides. The method is platform dependent. flock is not a tool but just a variable name which means fileLock. That's it.

PR #81, opened 2017-08-03
Add version for groovyserver support

dankwart-de:
Add autodownload for groovyserver.

Add a new env variable for groovy server.


RoddyCore/src/de/dkfz/roddy/Roddy.java
```diff
@@ -217,7 +217,11 @@ public static void main(String[] args) {
             mainStarted = true;
             startup(args);
         } catch (Exception e) {
-            e.printStackTrace();
+            // I did not want to put in a dependency to another library. This special type of
+            // exception is thrown by groovyclient (groovyserver project). As it is not known
+            // to the JVM in this case, we'll filter it by its name.
+            if (!e.getClass().getName().endsWith("SystemExitException"))
+                e.printStackTrace();
```
vinjana: Sollte dass nicht 

```
} catch (SystemExitException e) {
   e.printStackTrace();
}
```
sein? Warum printStackTrace(), wenn die Exception unwichtig ist?

dist/bin/current/helperScripts/setupRuntimeEnvironment.sh
```diff
@@ -58,9 +58,9 @@ else
   fi
 fi
 
->&2 echo $(readlink -f which java)
->&2 echo $(readlink -f which javac)
->&2 echo $(readlink -f which groovy)
+>&2 echo $(readlink -f $(which java))
```
vinjana: Is also in my pull request :-)

dist/bin/current/roddy.sh
```diff
@@ -129,7 +129,23 @@ elif [[ "$parm1" == "createworkflow" ]]; then
 fi
 
 IFS=""
-#[[ $RMIPORT != "" ]] && export DBG_OPTS="-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005" && echo "Opened up rmi debugger port"
-java $DBG_OPTS -cp .:$libraries:${RODDY_BINARY} de.dkfz.roddy.Roddy $*
-# -verbose:class
+export RGROOVY_LIB=`readlink -f ${RODDY_BINARY_DIR}/lib/groovy*.jar`
+
+# Check for groovyserv. If it does not exist, try to download it. If it still fails, start Roddy without it.
+caller=java
```
vinjana: What if someone is behind a firewall and gets long timeouts on every startup, because everytime the scripts tries to download groovyserv?

dist/bin/current/roddy.sh
```diff
@@ -129,7 +129,23 @@ elif [[ "$parm1" == "createworkflow" ]]; then
 fi
 
 IFS=""
-#[[ $RMIPORT != "" ]] && export DBG_OPTS="-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005" && echo "Opened up rmi debugger port"
-java $DBG_OPTS -cp .:$libraries:${RODDY_BINARY} de.dkfz.roddy.Roddy $*
-# -verbose:class
+export RGROOVY_LIB=`readlink -f ${RODDY_BINARY_DIR}/lib/groovy*.jar`
+
+# Check for groovyserv. If it does not exist, try to download it. If it still fails, start Roddy without it.
```
vinjana: What if someone is behind a firewall and gets long timeouts on every startup, because everytime the scripts tries to download groovyserv?

dist/bin/current/roddy.sh
```diff
@@ -129,7 +129,23 @@ elif [[ "$parm1" == "createworkflow" ]]; then
...
+caller=java
+if [[ ! -d dist/runtime/groovyserv* && ! -f dist/runtime/gservforbidden ]]; then
+	mkdir -p dist/runtime/ 2>/dev/null
+	(cd dist/runtime && wget -S $(curl -s https://kobo.github.io/groovyserv/download.html | grep groovyserv | grep bin.zip | sed 's/href=/\n/g' | sed 's/>/\n/g' | sed 's/</\n/g' | sed 's/"//g'  | grep .zip))
```
vinjana: Somewhere it should be documented how to install groovyserv, if the download here fails.
Also, better than just inlining complex bash code in an "if", write a download function and call it here.

dist/bin/current/roddy.sh
```diff
@@ -129,7 +129,23 @@ elif [[ "$parm1" == "createworkflow" ]]; then
 fi
 
 IFS=""
-#[[ $RMIPORT != "" ]] && export DBG_OPTS="-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005" && echo "Opened up rmi debugger port"
-java $DBG_OPTS -cp .:$libraries:${RODDY_BINARY} de.dkfz.roddy.Roddy $*
-# -verbose:class
+export RGROOVY_LIB=`readlink -f ${RODDY_BINARY_DIR}/lib/groovy*.jar`
```
vinjana: Actually, the whole setup should not be inlined but put into functions.

RoddyCore/src/de/dkfz/roddy/Roddy.java
dankwart-de: No, here I'll ignore the SystemExitException (see the negation). As I wrote in the comment, I did not want to add another library dependency.

dist/bin/current/roddy.sh
dankwart-de: Good point.

dist/bin/current/roddy.sh
dankwart-de: That's why the gservforbidden file will be created. If it exists, the groovyserver setup won't be touched anymore.

dist/bin/current/roddy.sh
dankwart-de: Ok, I'll put that into RTD.

dankwart-de: Done

dist/bin/current/roddy.sh
vinjana: Can you write an error/warning message that describes what to do (where to put the gservforbidden-file with what content), if a timeout occurs?

RoddyCore/src/de/dkfz/roddy/Roddy.java
vinjana: Sorry, did not really understand your code comment.

RoddyCore/src/de/dkfz/roddy/tools/RuntimeTools.groovy
```diff
@@ -51,6 +55,17 @@ public final class RuntimeTools {
     }
 
     public static File getGroovyLibrary() {
-        return new File(System.getProperty("java.class.path").split("[:]").find { new File(it).name.startsWith("groovy") })
+        // Try to get Groovy from the environment. This is needed for groovyserv.
+        // If it is not working get it from the classpath.
+        logger.rare(([""] + System.getenv().collect { String k, String v -> "${k}=${v}" }.join("\n") + [""]).flatten().join("\n"))
+        if (System.getenv().containsKey("RGROOVY_LIB")) {
```
vinjana: 'R'? What does that mean? Roddy? Why not then `RODDY_GROOVY_LIB`?

dist/bin/current/helperScripts/downloadFunctions.sh
```diff
@@ -0,0 +1,27 @@
...
+  # Check for groovyserv. If it does not exist, try to download it. If it still fails, start Roddy without it.
+  local callerBinary=java
+  local roddyApplicationDirectory=${1-""}
+  local runtimeFolder=${roddyApplicationDirectory}/dist/runtime
+  local forbiddenFile=${runtimeFolder}/gservforbidden
+
+  if [[ ! -d $(find ${runtimeFolder} -maxdepth 1 -mindepth 1 -type d -name "groovys*") && ! -f ${forbiddenFile} ]]; then
+
+    mkdir -p dist/runtime/ 2>/dev/null
+    (cd ${runtimeFolder} && wget -S $(curl -s https://kobo.github.io/groovyserv/download.html | grep groovyserv | grep bin.zip | sed 's/href=/\n/g' | sed 's/>/\n/g' | sed 's/</\n/g' | sed 's/"//g'  | grep .zip)) &> /dev/null
```
vinjana: Do we have pipefail and errexit turned on?

dist/bin/current/roddy.sh
```diff
@@ -129,7 +126,23 @@ elif [[ "$parm1" == "createworkflow" ]]; then
...
+debuggerSettings=""
+[[ -n ${DEBUG_ROODY} ]] && debuggerSettings=-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005
+
+caller=$(checkAndDownloadGroovyServ "${RODDY_DIRECTORY}")
+
+echo "RGL: "$RGROOVY_LIB
+
+if [[ ${caller} == java ]]; then
+  echo "Using Java to start Roddy"
+	${caller} ${debuggerSettings} -cp .:$libraries:${RODDY_BINARY} de.dkfz.roddy.Roddy $*
```
vinjana: What about the other Java options? In particular -Xms64m -Xmx512m? What about user JAVA_OPTS?

These are set in the top-level roddy.sh and should be used!

vinjana: 2nd: How does GroovyServ Java options? Also with GroovyServ we don't want exploding memory consumption (1/4 of the memory by default!). In the end, Roddy runtime is not only limited by Java startup time, but also by remote accesses.

RoddyCore/src/de/dkfz/roddy/tools/RuntimeTools.groovy
dankwart-de: done

dist/bin/current/roddy.sh
dankwart-de: JAVA_OPTS are automatically used by groovyserver. Just checked it, it's hidden in the .go source files.

RoddyCore/src/de/dkfz/roddy/Roddy.java
dankwart-de: Improved it.

dist/bin/current/roddy.sh
dankwart-de: Renamed the variable, created a method with a clear name.

dist/bin/current/helperScripts/downloadFunctions.sh
dankwart-de: Not yet, will do it.

dist/bin/current/roddy.sh
```diff
@@ -129,7 +128,25 @@ elif [[ "$parm1" == "createworkflow" ]]; then
...
-java $DBG_OPTS -cp .:$libraries:${RODDY_BINARY} de.dkfz.roddy.Roddy $*
-# -verbose:class
+export RODDIES_GROOVYLIB_PATH=`readlink -f ${RODDY_BINARY_DIR}/lib/groovy*.jar`
+
+debuggerSettings=""
+[[ -n ${DEBUG_ROODY} ]] && debuggerSettings=-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005
+
+caller=$(checkAndDownloadGroovyServ "${RODDY_DIRECTORY}")
+
+echo "RGL: "$RODDIES_GROOVYLIB_PATH
```
vinjana: Unconventional to use the Genitive here. Think of grepping for all Roddy variables! Would be harder like this.

dist/bin/current/helperScripts/downloadFunctions.sh
vinjana: Thanks!

dist/bin/current/helperScripts/downloadFunctions.sh
```diff
@@ -0,0 +1,27 @@
...
+
+    if [[ $? -ne 0 ]]; then
+      touch ${forbiddenFile}
```
vinjana: What about issuing a warning to the user via stderr? I know, most users will forget after some time, but some might not. It'd give the feeling that Roddy is not a blackbox, without overloading the user.

Alternatively/Additionally, you may add a warning that the forbiddenfile is used and no groovyserv is used.

dist/bin/current/helperScripts/downloadFunctions.sh
```diff
@@ -0,0 +1,27 @@
...
+    else
+      (cd ${runtimeFolder} && unzip groovyser*.zip && rm groovyser*.zip*)
+      [[ $? -ne 0 ]] && touch ${forbiddenFile}
```
vinjana: That's a different kind of error. If the download worked but the unzip did not, more is wrong than just the network. Just the question: Would that not warrant an error?

PR #82, opened 2017-08-03
Fix bejob and added md5 and script fixes

vinjana:
You were right. I forgot this pull request.

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/config/RecursiveOverridableMapContainer.java
+++ b/RoddyCore/src/de/dkfz/roddy/config/RecursiveOverridableMapContainer.java
@@ -143,6 +143,10 @@ public class RecursiveOverridableMapContainer<K, V extends RecursiveOverridableM
         return _getValueUnchecked(id);
     }
 
+    /**
+     * @throws RuntimeException if a value is requested that does not exist.
+     * @param id    Configuration value to return.
+     */
     public V getValue(String id) {
         if (!hasValue(id))
             throw new RuntimeException("Value " + id + " could not be found in containers with id " + id);
--- a/dist/bin/current/helperScripts/compileRoddyPlugin.sh
+++ b/dist/bin/current/helperScripts/compileRoddyPlugin.sh
@@ -4,13 +4,13 @@ set -e
 increasebuildonly=${increasebuildonly-false}
 
 # Check for configured plugin directories. Set empty if no dirs are configured.
-pluginDirectories=`grep pluginDirectories ${customconfigfile}`
-[[ -z "$pluginDirectories" ]] && pluginDirectories="pluginDirectories="
+pluginDirectories=`grepFromConfigFile pluginDirectories $customconfigfile`
+pluginDirectories="pluginDirectories=$pluginDirectories"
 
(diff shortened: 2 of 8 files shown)


dist/bin/current/helperScripts/compileRoddyPlugin.sh
```diff
@@ -4,13 +4,13 @@ set -e
 increasebuildonly=${increasebuildonly-false}
 
 # Check for configured plugin directories. Set empty if no dirs are configured.
-pluginDirectories=`grep pluginDirectories ${customconfigfile}`
-[[ -z "$pluginDirectories" ]] && pluginDirectories="pluginDirectories="
+pluginDirectories=`grepFromConfigFile pluginDirectories $customconfigfile`
```
vinjana: I changed the line and in the course of editing removed them, apparently. As they are meaningless in this context I could leave them away, but if you wish I add them again.

RoddyCore/src/de/dkfz/roddy/config/RecursiveOverridableMapContainer.java
```diff
@@ -143,6 +143,10 @@ public V getValue(String id, V defaultValue) {
         return _getValueUnchecked(id);
     }
 
+    /**
+     * @throws RuntimeException if a value is requested that does not exist.
+     * @param id    Configuration value to return.
+     */
```
vinjana: 1. Apparently, it is good practice to reuse official/existing core exceptions:
https://stackoverflow.com/questions/10016550/java-best-practices-when-throwing-exceptions-throwing-core-java-exceptions
2. There is https://docs.oracle.com/javase/7/docs/api/javax/xml/datatype/DatatypeConfigurationException.html, but the "Datatype..." does not fit. I would not use it here.
3. Maybe https://docs.oracle.com/javase/7/docs/api/java/text/ParseException.html, but I would not use it, because the configuration tree is not only build by parsing.

That's what I could find. I would make a new ConfigurationException.

vinjana: 1. Apparently, it is good practice to reuse official/existing core exceptions:
https://stackoverflow.com/questions/10016550/java-best-practices-when-throwing-exceptions-throwing-core-java-exceptions
2. There is https://docs.oracle.com/javase/7/docs/api/javax/xml/datatype/DatatypeConfigurationException.html, but the "Datatype..." does not fit. I would not use it here.
3. Maybe https://docs.oracle.com/javase/7/docs/api/java/text/ParseException.html, but I would not use it, because the configuration tree is not only build by parsing.

That's what I could find. I would make a new ConfigurationException.

dankwart-de: Ok, that sounds reasonable. Leave it for now, but I'll create an issue for it.

dist/bin/current/helperScripts/compileRoddyPlugin.sh
dankwart-de: Was only a question :D just leave it.

PR #87, opened 2017-08-09
Fixes for 2.4

vinjana:
Some changes to fix Roddy, adapt its code to the current BatchEuphoria and get the environment-script loading support running.


RoddyCore/src/de/dkfz/roddy/config/converters/BashConverter.groovy
```diff
@@ -37,22 +37,31 @@ class BashConverter extends ConfigurationConverter {
...
+        })
+
         StringBuilder text = createNewDocumentStringBuilder(context, cfg)
 
         text << appendConfigurationValues(context, cfg)
 
         text << appendConfigurationValueBundles(context, cfg)
 
-        text << appendToolEntries(context, cfg)
+        // text << appendToolEntries(context, cfg)
```
dankwart-de: Remove it.

RoddyCore/src/de/dkfz/roddy/execution/io/LocalExecutionService.groovy
```diff
@@ -72,6 +74,31 @@ public class LocalExecutionService extends ExecutionService {
         return outputStream;
     }
 
+    /**
```
dankwart-de: I don't think, this is the right place for it. Please consider to move it to RuntimeService or ExecutionContext.

dankwart-de: Also, there is a readfile method (or it has another name) in the executionservice classes. 
Why do you have the getJobStateInfoLine here but the getJobStateInfoLineForJob in the SSH service?
I'd rather like to see all the code in one class. Maybe JobStateLineHelper?

RoddyCore/src/de/dkfz/roddy/execution/io/SSHExecutionService.groovy
```diff
@@ -866,4 +871,43 @@ class SSHExecutionService extends RemoteExecutionService {
...
 
+    /**
+     * Stores a new job jobState info to an execution contexts job jobState log file.
+     *
+     * @param job
+     */
+    private static String jobStateInfoLine(String jobId, String code, String millis) {
+        return String.format("%s:%s:%s", jobId, code, millis)
+    }
+    private static String getJobStateInfoLineForJob(BEJob job) {
```
dankwart-de: Why is this method necessary? It looks partly like the method in LocalExecutionService? Maybe you can clarify it?

RoddyCore/src/de/dkfz/roddy/execution/io/LocalExecutionService.groovy
vinjana: The reason this is here, is that the jobstate logfile has different content when jobs are executed remotely on a cluster or locally (directly) and that I wanted to keep the original behaviour. 

But, I can also put into the RuntimeService. Nevertheless, there will be two versions, one for local job execution and one for job execution on the cluster. The alternative is to break the existing way it is done.

RoddyCore/src/de/dkfz/roddy/execution/io/SSHExecutionService.groovy
vinjana: This is to keep the old behaviour. See previous commend.

RoddyCore/src/de/dkfz/roddy/execution/io/LocalExecutionService.groovy
dankwart-de: Ok, then leave it for now and we better discuss it.

RoddyCore/src/de/dkfz/roddy/execution/jobs/Job.groovy
```diff
@@ -333,6 +338,100 @@ class Job extends BEJob<BEJob, JobResult> {
...
+
+    /**
+     * Stores a new job jobState info to an execution contexts job jobState log file.
+     *
+     * @param job
+     */
+    void appendToJobStateLogfile(BatchEuphoriaJobManager jobManager, ExecutionContext executionContext, BEJobResult res, OutputStream out = null) {
+        // Ugly, but works. Java/Groovy's missing runtime dispatch on method arguments sucks.
+        if (jobManager instanceof ClusterJobManager) {
+            appendToJobStateLogfileForCluster(executionContext, res)
```
dankwart-de: I don't like the idea here, that we need instanceof. This should be handled by Roddys ExecutionService (Local/SSH). There is the writeLine method or so. Or a pattern like that for Strategy or Visitor depending on the Roddy execution service might be good?

RoddyCore/src/de/dkfz/roddy/execution/jobs/Job.groovy
```diff
@@ -333,6 +338,100 @@ class Job extends BEJob<BEJob, JobResult> {
...
+    void appendToJobStateLogfileForCluster(ExecutionContext executionContext, BEJobResult res) {
+        if (res.wasExecuted) {
+            def job = res.command.getJob()
+            String jobInfoLine = getJobStateInfoLineForClusterJob(job)
+            if (jobInfoLine != null)
+                FileSystemAccessProvider.getInstance().appendLineToFile(true, executionContext.getRuntimeService().getNameOfJobStateLogFile(executionContext), jobInfoLine, false)
+        }
+    }
+
+    static String getJobStateInfoLineForClusterJob(BEJob job) {
```
dankwart-de: Oh that took me some time to see what happens :D 
Maybe some comments would help?
Maybe getJobState... is the wrong name? AssembleJobStateInfoLine?

RoddyCore/src/de/dkfz/roddy/execution/jobs/Job.groovy
```diff
@@ -333,6 +338,100 @@ class Job extends BEJob<BEJob, JobResult> {
...
+    private static String jobStateInfoLine(String jobId, String code, String millis) {
+        return String.format("%s:%s:%s", jobId, code, millis)
+    }
+
+
+    void appendToJobStateLogfileForCluster(ExecutionContext executionContext, BEJobResult res) {
+        if (res.wasExecuted) {
+            def job = res.command.getJob()
+            String jobInfoLine = getJobStateInfoLineForClusterJob(job)
+            if (jobInfoLine != null)
```
dankwart-de: Could we merge this with the method below?

RoddyCore/src/de/dkfz/roddy/execution/jobs/JobManager.groovy
```diff
@@ -46,8 +46,8 @@ class JobManager {
         jobManager.runJob(job, runDummy)
     }
 
-    BEJobDependencyID createJobDependencyID(Job job, String jobResult) {
-        jobManager.createJobDependencyID(job, jobResult)
+    BEJobID createJobDependencyID(Job job, String jobResult) {
+        jobManager.createID(job, jobResult)
```
dankwart-de: createJobID?

RoddyCore/src/de/dkfz/roddy/execution/jobs/Job.groovy
vinjana: Yes, that's better.

RoddyCore/src/de/dkfz/roddy/execution/jobs/JobManager.groovy
vinjana: Yes, better!

RoddyCore/src/de/dkfz/roddy/execution/jobs/Job.groovy
vinjana: I don't think so, if we don't want to add more lines to the jobStateLogfile.

In principle I do not have any problems with adding new lines to the jobStateLogfile (but it needs to be communicated to the OTP crowd). The jobStateLogFile would then even more become an event log (submitted, started, stopped-with-exit-code, etc.).

RoddyCore/src/de/dkfz/roddy/execution/jobs/Job.groovy
vinjana: Neither do I. It is because Groovy does not support dynamic dispatch on more than the object itself (a.k.a. multiple dispatch). But I agree with your previous comment (!) that this code does not belong into ExecutionService. Conceptually, this code belongs here. 

I now better understand the other instanceOf cases that we have. They are there due to the missing features in the object system (or of object systems in general).

PR #97, opened 2017-09-06
Introduce submodules for plugins

dankwart-de:
Move DefaultPlugin and PluginBase to custom repos. Add them as submodules.


.gitmodules
```diff
@@ -0,0 +1,6 @@
+[submodule "dist/plugins/PluginBase"]
```
vinjana: Please add a remark to the developer docs, that `git submodules` something is required to work with Roddy.

PR #113, opened 2017-09-28
Add --additionalImports option (both lower and uppercase)

dankwart-de:


RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -314,6 +315,19 @@ class ProjectLoader {
     }
 
     /**
+     * If --additionalImports is used, try to attach all configuration objects to the currently loaded
```
vinjana: I don't understand the comment. Is that not something that should happen anyway unconditionally?

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -154,6 +154,8 @@ class ProjectLoader {
 
             loadProjectConfigurationOrFail(projectID, analysisID)
 
+            attachAdditionallyPassedConfigurationsOrFail()
```
vinjana: `orFail`? Returns `void` and does not obviously throw?!

dankwart-de:
Yes tried it out. But you are right.  Bother the comment and the fail is not well handled.

PR #114, opened 2017-09-28
Add check for i/o directory accessibility

dankwart-de:
As discussed, some preliminary checks for the in and outputdirectories should improve the application startup.

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
+++ b/RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
@@ -469,26 +469,34 @@ class ProjectLoader {
 
         // Earliest check for valid input and output directories. If they are not accessible or writeable.
         // Start with the input directory
-        for (File _dir = analysis.getInputBaseDirectory(); _dir; _dir = _dir.parentFile) {
-            if (!FileSystemAccessProvider.instance.isReadable(_dir) || !FileSystemAccessProvider.instance.isExecutable(_dir)) {
-                errors << "The input directory was not readable at path ${_dir}."
-                break
-            }
-        }
+        errors += checkDirForReadabilityAndExecutability(analysis.getInputBaseDirectory())
+        errors += checkDirForReadabilityAndExecutability(analysis.getOutputBaseDirectory())
 
-        for (File _dir = analysis.getOutputBaseDirectory(); _dir; _dir = _dir.parentFile) {
-            if (!FileSystemAccessProvider.instance.isReadable(_dir) || !FileSystemAccessProvider.instance.isExecutable(_dir))
-                if (!FileSystemAccessProvider.instance.isReadable(_dir)) {
-                    errors << "The output directory was not readable at path ${_dir}."
-                    break
-                }
-        }
-
+        // Out dir needs to be writable
         if (!FileSystemAccessProvider.instance.isWritable(analysis.getOutputBaseDirectory()))
             errors << "The output was not writeable at path ${analysis.getOutputBaseDirectory()}."
 
-        if (errors) {
-            throw new ProjectLoaderException((["There were errors in directory access checks:"] + errors).join("\t\n"))
+        if (!errors)
+            return
+
+        throw new ProjectLoaderException((["There were errors in directory access checks:"] + errors).join("\t\n"))
+    }
+
+    List<String> checkDirForReadabilityAndExecutability(File dirToCheck) {
+        List<String> errors = []
+        for (File _dir = dirToCheck; _dir; _dir = _dir.parentFile) {
+            boolean readable = FileSystemAccessProvider.instance.isReadable(_dir)
+            boolean executable = FileSystemAccessProvider.instance.isExecutable(_dir)
+            if (!readable || !executable) {
+                if (!readable && !executable)
+                    errors << "The output directory was neither readable nor executable at path ${_dir}."
+                else if (!readable)
+                    errors << "The output directory was not readable at path ${_dir}."
+                else if (!executable)
+                    errors << "The output directory was not executable at path ${_dir}."
+                break
+            }
         }
+        return errors
     }
 }
--- a/RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
+++ b/RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
@@ -465,20 +465,30 @@ class ProjectLoader {
             throw new ProjectLoaderException("There is no runtime service class set for the selected analysis. This has to be set in either the project configuration or the analysis configuration.");
         }
 
+        List<String> errors = []
+
         // Earliest check for valid input and output directories. If they are not accessible or writeable.
         // Start with the input directory
         for (File _dir = analysis.getInputBaseDirectory(); _dir; _dir = _dir.parentFile) {
-            if (!FileSystemAccessProvider.instance.isReadable(_dir))
-                throw new ProjectLoaderException("The input directory was not readable at path ${_dir}.")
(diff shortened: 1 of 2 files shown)


RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -457,8 +459,26 @@ class ProjectLoader {
...
-        } else if (analysis.getRuntimeService() == null) {
+        }
+
+        if (analysis.getRuntimeService() == null) {
             throw new ProjectLoaderException("There is no runtime service class set for the selected analysis. This has to be set in either the project configuration or the analysis configuration.");
         }
+
+        // Earliest check for valid input and output directories. If they are not accessible or writeable.
+        // Start with the input directory
+        for (File _dir = analysis.getInputBaseDirectory(); _dir; _dir = _dir.parentFile) {
```
vinjana: Is there no execution context available yet (addErrorEntry)?

vinjana: BTW Does the addErrorEntry not make more sense in the `Roddy` class? We are dealing with globally fatal errors, so why not put it into the top-level? If so this should be refactored before release of 2.4-final.

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -457,8 +459,26 @@ class ProjectLoader {
...
+            if (!FileSystemAccessProvider.instance.isReadable(_dir))
+                throw new ProjectLoaderException("The input directory was not readable at path ${_dir}.")
+        }
+
+        for (File _dir = analysis.getOutputBaseDirectory(); _dir; _dir = _dir.parentFile) {
+            if (!FileSystemAccessProvider.instance.isReadable(_dir))
```
vinjana: The output directory does not need to be writable, right? Some files may go into existing (writable) subdirectories, right?

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
dankwart-de: There is no context available at this point.

It might make sense to have an addErrorEntry class but errors on startup are hardly collected and normally lead to a direct application abort.

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
dankwart-de: It must be because the dataset folders will be created here, also the execution cache dir and the execution list. If they were already create ok, we could skip it. But then it is in the context and I think it might already be checked there.

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
dankwart-de: Of course not a class but a method.

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
vinjana: So, you should add a test for writability at the terminal output directory then.

BTW: The error message is wrong (says input directory).

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -457,8 +459,36 @@ class ProjectLoader {
...
+        // Start with the input directory
+        for (File _dir = analysis.getInputBaseDirectory(); _dir; _dir = _dir.parentFile) {
+            if (!FileSystemAccessProvider.instance.isReadable(_dir) || !FileSystemAccessProvider.instance.isExecutable(_dir)) {
+                errors << "The input directory was not readable at path ${_dir}."
+                break
+            }
+        }
+
+        for (File _dir = analysis.getOutputBaseDirectory(); _dir; _dir = _dir.parentFile) {
+            if (!FileSystemAccessProvider.instance.isReadable(_dir) || !FileSystemAccessProvider.instance.isExecutable(_dir))
```
vinjana: All input directory elements need to be executable, but the last additionally needs to be readable.

All output directory elements need to be executable, but the last additionally needs to be readable and writable.

This is not exactly what is tested here (although the use case, that the intermediate paths are only executable but not readable should be rare; I saw that mostly for paths exposed via an HTTP server.)

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -457,8 +459,36 @@ class ProjectLoader {
...
+                if (!FileSystemAccessProvider.instance.isReadable(_dir)) {
```
vinjana: You only report, if it is not readable? You test for executability but don't report, if it not executable?

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -457,8 +459,36 @@ class ProjectLoader {
...
+                    errors << "The output directory was not readable at path ${_dir}."
+                    break
+                }
+        }
+
+        if (!FileSystemAccessProvider.instance.isWritable(analysis.getOutputBaseDirectory()))
```
vinjana: Can you keep the input dir tests together?

dankwart-de: Don't exactly know what you mean here.

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -469,26 +469,34 @@ class ProjectLoader {
...
+    List<String> checkDirForReadabilityAndExecutability(File dirToCheck) {
+        List<String> errors = []
+        for (File _dir = dirToCheck; _dir; _dir = _dir.parentFile) {
+            boolean readable = FileSystemAccessProvider.instance.isReadable(_dir)
+            boolean executable = FileSystemAccessProvider.instance.isExecutable(_dir)
+            if (!readable || !executable) {
+                if (!readable && !executable)
+                    errors << "The output directory was neither readable nor executable at path ${_dir}."
+                else if (!readable)
+                    errors << "The output directory was not readable at path ${_dir}."
```
vinjana: input directory?

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -469,26 +469,34 @@ class ProjectLoader {
...
-        }
-
-        for (File _dir = analysis.getOutputBaseDirectory(); _dir; _dir = _dir.parentFile) {
-            if (!FileSystemAccessProvider.instance.isReadable(_dir) || !FileSystemAccessProvider.instance.isExecutable(_dir))
-                if (!FileSystemAccessProvider.instance.isReadable(_dir)) {
-                    errors << "The output directory was not readable at path ${_dir}."
-                    break
-                }
-        }
+        errors += checkDirForReadabilityAndExecutability(analysis.getInputBaseDirectory())
```
vinjana: O.k. you don't want to discriminate between inner path elements (executability-only is fine) and terminal path elements (readable and executable is necessary (and writable for output dir). Fine.

dankwart-de: Don't know, if it is really necessary. Let's just see, how it behaves? We could also create a flag to ignore this.

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
dankwart-de: Done

PR #122, opened 2017-10-05
Add strict config duplicate handling, fixes

dankwart-de:
Fix output for configuration loader messages

Strict mode for configuration ids enforces clean directories as no
duplicate ids are allowed anymore.


dankwart-de:
Clean here means, that there must not be multiple configuration files using the same configuration ids!

vinjana:
So it has nothing to do with directories?

PR #128, opened 2017-10-10
Fix plugin directory check

vinjana:


RoddyCore/src/de/dkfz/roddy/plugins/LibrariesFactory.groovy
```diff
@@ -308,48 +308,47 @@ public class LibrariesFactory extends Initializable {
         return file.exists() && file.isDirectory() && file.canRead() && file.canExecute()
     }
 
-    static PluginType determinePluginType(File directory, Map<String, List<String>> mapOfErrors = [:]) {
-        logger.postRareInfo("  Parsing plugin folder: ${directory}");
+    static PluginType determinePluginType(File file, Map<String, List<String>> mapOfErrors = [:]) {
```
dankwart-de: Why did you rename it to file? It is supposed to be a directory (see first check below)

RoddyCore/src/de/dkfz/roddy/plugins/PluginInfo.java
```diff
@@ -99,21 +104,25 @@ public PluginInfo(String name, File zipFile, File directory, File developmentDir
...
-                        errors.add("Ignore directory " + file.getAbsolutePath() + "; It is not a valid tools directory.");
+                    PosixFileAttributes attr;
+                    try {
+                        attr = Files.readAttributes(file.toPath(), PosixFileAttributes.class);
+                    } catch (IOException ex) {
+                        errors.add("An IOException occurred while accessing '" + file.getAbsolutePath() + "': " + ex.getMessage());
+                        continue;
+                    }
+
+                    if (attr.isRegularFile() || file.isHidden()) {
```
dankwart-de: I don't know if this check is valid or behaves like it should. The former code checked, if "file" is  not a directory. Here you turn it around and check if it is a regular file. What about links etc.?

RoddyCore/src/de/dkfz/roddy/plugins/PluginInfo.java
```diff
@@ -99,21 +104,25 @@ public PluginInfo(String name, File zipFile, File directory, File developmentDir
 
     protected void fillListOfToolDirectories() {
         File toolsBaseDir = null;
-        try {
-            toolsBaseDir = getToolsDirectory();
+        toolsBaseDir = getToolsDirectory();
 
-            if (toolsBaseDir != null && toolsBaseDir.exists() && toolsBaseDir.isDirectory()) { //Search through the default folders, if possible.
+        if (toolsBaseDir != null && toolsBaseDir.exists() && toolsBaseDir.isDirectory()) { //Search through the default folders, if possible.
```
dankwart-de: Maybe you could invert the if? Then we'd have one less level of intendation.

RoddyCore/src/de/dkfz/roddy/plugins/LibrariesFactory.groovy
vinjana: You're right. Changed it back.

RoddyCore/src/de/dkfz/roddy/plugins/PluginInfo.java
vinjana: Yes, I'll change it back. For some reason I remembered (incorrectly) that the access via PosixFileAttributes dereferences a symlink. But that is obviously wrong.

PR #134, opened 2017-10-26
Removed (almost) all source $CONFIG_FILE from scripts (outdated or no…

vinjana:
…t). Added throws ConfigurationError, where required.


RoddyCore/src/de/dkfz/roddy/knowledge/files/FileObject.java
```diff
@@ -26,7 +27,7 @@ public FileObject(ExecutionContext executionContext) {
         this.executionContext = executionContext;
     }
 
-    public abstract void runDefaultOperations();
+    public abstract void runDefaultOperations() throws ConfigurationError;
```
dankwart-de: Why does this actually throw a configuration error? I mean it is a BaseFile class.

Also we could think of getting rid of this. The mechanism was hardly used.

dist/bin/current/helperScripts/findPluginLibraries.groovy
```diff
@@ -4,37 +4,88 @@
  * Distributed under the MIT License (license terms are at https://www.github.com/eilslabs/Roddy/LICENSE.txt).
  */
 
+
```
dankwart-de: you extended this a lot and this is good. But as it is Groovy, we could also use the Roddy code itself to find all this or not? At least, we should investigate it.

RoddyCore/src/de/dkfz/roddy/knowledge/files/FileObject.java
vinjana: I admit that I made many of the changes, just because the compiler required them. Given the sometimes convoluted structure of Roddy, it is possible that the derived class code has the side-effect of accessing the configuration and thus the possibility to throw a ConfigurationError. If we could prevent this, this would certainly improve the code!

dist/bin/current/helperScripts/findPluginLibraries.groovy
vinjana: Right.

PR #136, opened 2017-11-06
Add better listworkflows support. Reenable it.

dankwart-de:


RoddyCore/src/de/dkfz/roddy/client/cliclient/RoddyCLIClient.groovy
```diff
@@ -425,15 +426,30 @@ public class RoddyCLIClient {
...
             configText << prefix << " #FBLUE#Available analyses:#CLEAR#" << separator
             int i = 0;
-            for (String iccAna in pti.icc.listOfAnalyses) {
-                String[] split = iccAna.split("::");
-                configText << prefix << "   " << i << ": ID=[" << split[-2] << "], Workflow=[" << split[-1] << "]" << separator;
-                i++;
+            int longestAnalysisID = 0
+            int longestAnalysisSrcID = 0
+            int longestPluginID = 0
+            List<List<String>> splitted = pti.icc.listOfAnalyses.sort().collect {
```
vinjana: This is again the '::' separator for the analysis ID. Did you consider introducing a dedicated class, that handles this, instead of coding information in a string? This could be a simple value class, but would simplify a lot of client code (this site here is likely not the only one, we worked at such code for the filename patterns).

Parsing is notoriously hard and if things are not checked (like using array indexing to access non-existing parse results) things can break.

dankwart-de: Yes I thought about it. The fact, why I did not introduce it, is that we only use it rarely and here it also works slightly different than in the other case.

vinjana: So you think, making this a class is not worth the effort? I remember that there were bugs in this code at the other site.

dankwart-de: No, if you consider it a bug, then we can of course fix it. Better one more fix than bad code :D. In that case it just did not match.

PR #144, opened 2017-11-10
Allow Roddy to skip full job manager init for most start modes

dankwart-de:
e.g. testrun/testrerun etc. do not need to init a cluster job manager.
In these cases, the direct sync job manager will be used.

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/Roddy.java
+++ b/RoddyCore/src/de/dkfz/roddy/Roddy.java
@@ -7,6 +7,7 @@
 package de.dkfz.roddy;
 
 import com.btr.proxy.search.ProxySearch;
+import de.dkfz.roddy.client.RoddyStartupModeScopes;
 import de.dkfz.roddy.client.RoddyStartupModes;
 import de.dkfz.roddy.client.RoddyStartupOptions;
 import de.dkfz.roddy.client.cliclient.CommandLineCall;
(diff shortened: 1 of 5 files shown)


RoddyCore/src/de/dkfz/roddy/Roddy.java
```diff
@@ -276,17 +278,19 @@ private static void startup(String[] args) {
...
-            time("parsemode");
-            performCLIExit(clc.startupMode);
-        } else {
+        if (!initializeServices(clc.startupMode)) {
             logger.postAlwaysInfo("Could not initialize services. Roddy will not execute jobs.");
             performCLIExit(clc.startupMode, 1);
         }
+
+        if (!jobExecutionRequirementsFulfilled())
+            exit(1);
```
vinjana: No message here?

RoddyCore/src/de/dkfz/roddy/client/RoddyStartupModeScopes.java
```diff
@@ -19,16 +19,30 @@
...
      */
-    public static final int SCOPE_FULL = 0x20;
+    SCOPE_FULL(true),
 
+    /**
+     * This scope starts all modes and also initializes the Job Manager
+     */
+    SCOPE_FULL_WITHJOBMANAGER(true);
+
+    public final boolean needsFullInit;
```
vinjana: I see a conceptual problem here. 'full' kind of implies that the JM is also initialized as requested. As we have it in the moment, different configuration toggles/alternatives blend in this multifunctional enum. So the current design fails to implement the single responsibility principle.

We could initialize each start-up mode with just a set of toggles that indicate, whether or not specific features should be turned on or of (or independent enums, if its not just toggles). Would that make sense?

RoddyCore/src/de/dkfz/roddy/Roddy.java
dankwart-de: Errors inside :D

RoddyCore/src/de/dkfz/roddy/client/RoddyStartupModeScopes.java
dankwart-de: As discussed, I'll leave it.

dankwart-de:
Ok, as discussed some minutes ago, I'd leave it now and merge it.

vinjana:
Okidoki.

Issue #106, opened 2017-09-25
Make the blacklist for configuration values configurable (ini, cvalue)


vinjana:
`List<String> valueBlacklist = [ConfigurationConstants.CVALUE_PLACEHOLDER_RODDY_JOBID_RAW, 'PWD', Constants.PID_CAP, Constants.PID`

vinjana:
RoddyCore/src/de/dkfz/roddy/config/validation/ScriptValidator.groovy
RoddyCore/src/de/dkfz/roddy/config/converters/BashConverter.groovy

dankwart-de:
Solved / Ignored. We removed most of the blacklisted values, as we changed Roddys configuration value sort algorithm.

PR #146, opened 2017-11-13
Add blacklist cval for ScriptValidator/BashConverter

dankwart-de:
Both were inline in the code. Now they are in the default.xml! So we'll
see, where we need to put them as well.

See #106

I think it would be good to have defaults somewhere. Where to put them? Roddy class?


RoddyCore/src/de/dkfz/roddy/config/converters/BashConverter.groovy
```diff
@@ -134,81 +134,69 @@ class BashConverter extends ConfigurationConverter {
...
+        List<ConfigurationValue> values = cfg.getConfigurationValues().getAllValuesAsList()
+        Map<String, ConfigurationValue> listOfUnsortedValues = values.collectEntries { [it.id, it] }
+        Map<String, ConfigurationValue> listOfSortedValues = new LinkedHashMap<String, ConfigurationValue>()
 
-        for (ConfigurationValue cv in values) {
-            listOfUnsortedValues[cv.id] = cv;
-        }
+        // Blacklist of values which will be ignored for the dependency resolution. E.g. values, which will be available later on
+        // in the runtime config / job.
+        List<String> valueBlacklist = cfg.configurationValues.get("blacklistForCfgToRuntimeCfgConversion").toStringList()
```
vinjana: Not really a nice configuration option name.

RoddyCore/src/de/dkfz/roddy/config/converters/BashConverter.groovy
```diff
@@ -134,81 +134,69 @@ class BashConverter extends ConfigurationConverter {
...
+        return text
     }
 
     /**
      *
-     * @param cfg    Configuration object. Basically a tree of configuration values that may additionally contain cross-references in the values.
+     * @param cfg Configuration object. Basically a tree of configuration values that may additionally contain cross-references in the values.
      * @return
      */
     Map<String, ConfigurationValue> getConfigurationValuesSortedByDependencies(Configuration cfg) {
```
vinjana: Still a pretty long method. Can it be broken down and made more descriptive? Why put everything into a single for-loop? Maybe use prefiltering to get rid of the useless stuff first?

```groovy
def boolean isValidationRule(String configurationName) {
   configurationName.contains("cfgValidationRule")
}

def boolean isComment(String configurationValue) {
   cv.value != null && cv.value.trim().startsWith("#")
}

values.findAll { !isValidationRule(it.id) }.findAll { it.value != null && !isComment(it.value) }
```

... then the rest.

RoddyCore/src/de/dkfz/roddy/config/converters/BashConverter.groovy
```diff
@@ -134,81 +134,69 @@ class BashConverter extends ConfigurationConverter {
...
-        def values = cfg.getConfigurationValues().getAllValuesAsList();
-        Map<String, ConfigurationValue> listOfUnsortedValues = [:]
-        def listOfSortedValues = new LinkedHashMap<String, ConfigurationValue>();
+        List<ConfigurationValue> values = cfg.getConfigurationValues().getAllValuesAsList()
+        Map<String, ConfigurationValue> listOfUnsortedValues = values.collectEntries { [it.id, it] }
```
vinjana: Can you express somehow the sorting order in the name of the variable? Can you describe the algorithm in a comment? Do I see this correctly that here you generate a list with a partial order on the number of dependencies?

RoddyCore/src/de/dkfz/roddy/config/RecursiveOverridableMapContainerForConfigurationValues.java
```diff
@@ -74,6 +76,18 @@ public boolean getBoolean(String id, boolean b) {
         }
     }
 
+    public List<String> getList(String id) {
+        return getList(id, ",");
+    }
+
+    public List<String> getList(String id, String separator) {
+        try {
+            return Arrays.asList(getValue(id).toString().split("[" + separator + "]"));
```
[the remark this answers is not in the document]
dankwart-de: Currently, it is all trimmed, but I will also include code to trim every value.

vinjana:
If this is done, please merge it.

RoddyCore/src/de/dkfz/roddy/config/converters/BashConverter.groovy
```diff
@@ -134,91 +134,88 @@ class BashConverter extends ConfigurationConverter {
...
+        StringBuilder text = new StringBuilder()
 
         //Set a path if necessary.
         text << separator << new BashCommandSet().getSetPathCommand()
-        return text;
+        return text
     }
 
     /**
+     * @param cfg Configuration object. Basically a tree of configuration values that may additionally contain cross-references in the values.
```
vinjana: "Values with unresolved dependencies (i.e. variables that are not declared in the configuration tree) are put at the end. If `returnLeftOverCValuesOnly` is set, only the configuration values referencing undeclared variables are returned." ... or so.

vinjana: Actually, a cleaner interface would be a returning a pair of cvalues with resolved and unresolved references. Then in another small function, the two sets could be combined. The switch would be unnecessary. Furthermore, the name does not indicate, that the function either returns only unresolved or all variables.

dankwart-de: I am not really sure how to do that now. I changed the comment but I don't have a good idea for the interface. I'd leave it for now and create a new issue.

vinjana: As I said:

```groovy
Map<String, ConfigurationValue> getConfigurationValuesSortedByDependencies(Configuration cfg) {
   Tuple2<Map<String,ConfigurationValue>> sortedAndRest = getConfigurationValuesSortedByDependenciesAndUnresolvable(Configuration cfg)
  return sortedAndRest.first + sortedAndRest.second
}

Tuple2<Map<String,ConfigurationValue>> getConfigurationValuesSortedByDependencies(Configuration cfg) {
   // more or less what you have now
   return new Tuple2(listOfSortedValues, listOfUnsortedValues)
}
```

or so

dankwart-de:
Things are done, talked about it, merged it.

PR #147, opened 2017-11-13
Added RODDY_JOBNAME and fixed RODDY_QUEUE.

vinjana:


RoddyCore/RoddyCore.iml
```diff
@@ -14,5 +14,6 @@
     <orderEntry type="library" name="R Skeletons" level="application" />
     <orderEntry type="module" module-name="RoddyToolLib_main" />
     <orderEntry type="module" module-name="BatchEuphoria_main" />
+    <orderEntry type="library" name="roddyBundledLibraries" level="project" />
```
dankwart-de: I'm not sure. Didn't we agree on a global lib?

RoddyCore/src/de/dkfz/roddy/Roddy.java
```diff
@@ -589,8 +589,12 @@ private static void setDefaultRoddyJobIdVariable(RecursiveOverridableMapContaine
         configurationValues.add(new ConfigurationValue(CVALUE_PLACEHOLDER_RODDY_JOBID_RAW, "$" + Roddy.jobManager.getJobIdVariable()));
     }
 
+    private static void setDefaultRoddyJobNameVariable(RecursiveOverridableMapContainerForConfigurationValues configurationValues) {
```
dankwart-de: With all the (new) getters and setters, I think it might make sense to move all this to a custom class. Also we could have a RoddyStarter or so.. Might make things cleaner. Let's discuss it.

vinjana: Actually, I was also considering absorbing the single-line methods again into the only function they are called from. Generally, I think splitting the Roddy class may be useful, but I first want to understand which functions are there. The setters alone I would not consider enough for a class.

RoddyCore/RoddyCore.iml
vinjana: Sorry for this change. Yes we agreed on project-level library. This change reverts one that sneaked into a previous diff, because I was testing gradle.

PR #150, opened 2017-11-15
Gradle build of Roddy core

vinjana:
Separate zips for RoddyCore and RoddyEnv, the former containing /dist/bin/current + some addititional files, the latter containing the top-level roddy.sh + additional files and some dist/ directories as roddy environment.

Roddy seems to run with the fat (shadow) JAR and compilation integrates with RoddyToolLib and BE. No excessive tests of Roddy functions, yet.

To get (most) tests running there is also changes in the tests and how Roddy figures out directories.


vinjana:
There is still 2 tests failing. 
* NullPointerException from de.dkfz.roddy.config.FilenamePattern.fillValuesFromSourceFile in testFindFilenameFromSourcefilePatterns
* de.dkfz.roddy.config.converters.BashConverterTest#convertConfigurationValueToShellScriptLine

The latter test fails, because there is no jobManager set on Roddy and ConfigurationValue.toFile() dies.

RoddyCore/src/de/dkfz/roddy/tools/RuntimeTools.groovy
```diff
@@ -37,4 +48,20 @@ final class RuntimeTools {
         return new File(System.getProperty("user.dir"), "dist/bin/current")
     }
 
+    static File getGroovyLibrary() {
+        // Try to get Groovy from the environment. This is needed for groovyserv.
+        // If this is not working get it from the classpath.
+        logger.rare(([""] + System.getenv().collect { String k, String v -> "${k}=${v}" }.join("\n") + [""]).flatten().join("\n"))
```
dankwart-de: Will this work with the Gradle build, when no groovy library is e.g. in the lib folders?

vinjana: Tried it once and worked. I think I will try this also  out with my AQCWF integration tests.

PR #159, opened 2017-11-24
Fixed FilenamePattern.fillConfigurationVariables

dankwart-de:


RoddyCore/src/de/dkfz/roddy/config/FilenamePattern.java
```diff
@@ -183,10 +183,14 @@ String fillDirectories(String src, ExecutionContext context) {
      */
     String fillConfigurationVariables(String src, Configuration cfg) {
         RecursiveOverridableMapContainerForConfigurationValues configurationValues = cfg.getConfigurationValues();
-        while (src.contains(PLACEHOLDER_CVALUE)) {
+
+        for (boolean valueChanged = true; valueChanged && src.contains(PLACEHOLDER_CVALUE); ) {
```
vinjana: While-loop would be more expressive here.

PR #178, opened 2018-01-12
Alter roddy compile to use gradlew

dankwart-de:
Also fix a typo in console output

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
+++ b/RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
@@ -155,12 +155,15 @@ class ProjectLoader {
 
             checkAndPreparePremisesOrFail()
 
+            println()
             loadPluginsOrFail(projectID, analysisID)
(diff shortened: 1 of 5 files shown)


RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -155,12 +155,15 @@ class ProjectLoader {
 
             checkAndPreparePremisesOrFail()
 
+            println()
```
vinjana: Should be stderr!? Logger?!

The stdout is parsed by OTP.

dist/bin/current/helperScripts/compileRoddyBinary.sh
```diff
@@ -1,6 +1,5 @@
 #!/bin/bash
 
-#set -xuv
 set -e
 
 echo "Increasing build number and date to"
```
vinjana: Remove '&' in next statement.

PR #179, opened 2018-01-15
Fix roddy compile command

dankwart-de:
Remove content of dist/bin/current/lib first

Supress a message saying there is no content in lib folder

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/config/loader/ConfigurationFactory.groovy
+++ b/RoddyCore/src/de/dkfz/roddy/config/loader/ConfigurationFactory.groovy
@@ -324,12 +324,22 @@ class ConfigurationFactory {
         return loadConfiguration(icc)
     }
 
+    /**
+     * Will format a map to a two column table. The width of the first column will be calculated using the length of the
+     * longest element.
+     * @param map
+     * @param cntOfTabs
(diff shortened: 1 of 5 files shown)


RoddyCore/src/de/dkfz/roddy/config/loader/ConfigurationFactory.groovy
```diff
@@ -324,12 +324,22 @@ class ConfigurationFactory {
...
+     * @param map
+     * @param cntOfTabs
+     * @param tabSep
+     * @param closureForValue
+     * @return
+     */
+    @Deprecated
     @CompileDynamic
-    static List<String> convertMapToFormattedTable(Map map, int cntOfTabs, String tabSep, def clojureForValue) {
+    static List<String> convertMapToFormattedTable(Map map, int cntOfTabs, String tabSep, def closureForValue) {
```
vinjana: Not sure, but I think the type for closures is simply `Closure`. Maybe one could get rid of the `@CompileDynamic` here.

dankwart-de: Well that's how I learned it :D... but you're right, we can CompileStatic it.

PR #180, opened 2018-01-16
Fix getSourcefile in BaseFile class

dankwart-de:
A false parameter was passed within the method and this lead to a
ClassCastException

### DIFF
--- a/dist/plugins/TestPluginWithJarFile/src/de/dkfz/roddy/knowledge/examples/SimpleWorkflow.java
+++ b/dist/plugins/TestPluginWithJarFile/src/de/dkfz/roddy/knowledge/examples/SimpleWorkflow.java
@@ -9,6 +9,7 @@ package de.dkfz.roddy.knowledge.examples;
 
 import de.dkfz.roddy.core.ExecutionContext;
 import de.dkfz.roddy.core.Workflow;
+import de.dkfz.roddy.knowledge.files.BaseFile;
 import de.dkfz.roddy.knowledge.files.Tuple4;
 
 /**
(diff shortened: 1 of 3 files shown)


dist/plugins/TestPluginWithJarFile/src/de/dkfz/roddy/knowledge/examples/SimpleWorkflow.java
```diff
@@ -17,6 +18,7 @@
     @Override
     public boolean execute(ExecutionContext context) {
         SimpleRuntimeService srs = (SimpleRuntimeService) context.getRuntimeService();
+        BaseFile abc = getSourceFile(context, "/tmp/test.txt");
```
vinjana: I am not a fan of polluting /tmp. Roddy already puts so many files there that have to be cleaned up manually. We should consider making an issue for cleaning up /tmp after tests.

Here the file should be created in a dedicated subdirectory for /tmp.

BTW: `abc`?

dist/plugins/TestPluginWithJarFile/src/de/dkfz/roddy/knowledge/examples/SimpleWorkflowWithCorrectExecution.java
```diff
@@ -17,6 +18,7 @@
     @Override
     public boolean execute(ExecutionContext context) {
         SimpleRuntimeService srs = (SimpleRuntimeService) context.getRuntimeService();
+        BaseFile abc = getSourceFile(context, "/tmp/test.txt");
```
vinjana: dito

dist/plugins/TestPluginWithJarFile/src/de/dkfz/roddy/knowledge/examples/SimpleWorkflow.java
dankwart-de: This is really only a test workflow which is currently not used in any kind of automatic test.
The file also does not exist.

dist/plugins/TestPluginWithJarFile/src/de/dkfz/roddy/knowledge/examples/SimpleWorkflowWithCorrectExecution.java
dankwart-de: dito

PR #181, opened 2018-01-17
Modified gradle.build to clean all libraries (*.jar) from the dist/bi…

vinjana:
…n/current/libs before the `jar` target, i.e. before every library packing. Cleaning is also done for the `clean` target.


dankwart-de:
If you put it to gradle, you could also remove the rm lines from the compile script!

PR #185, opened 2018-01-19
Pass environment

vinjana:
1 line + many lines smaller refactorings

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/core/Analysis.java
+++ b/RoddyCore/src/de/dkfz/roddy/core/Analysis.java
@@ -200,7 +200,10 @@ public class Analysis {
         long creationCheckPoint = System.nanoTime();
 
         for (DataSet ds : selectedDatasets) {
-            if (level.isOrWasAllowedToSubmitJobs && !checkJobStartability(ds)) continue;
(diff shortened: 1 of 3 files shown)


vinjana:
PR fails yet, because wrong branch is referenced in .travis.yml. Need to wait for upstream changes and rerun test.

RoddyCore/src/de/dkfz/roddy/core/Analysis.java
```diff
@@ -222,7 +225,10 @@ public RuntimeService getRuntimeService() {
         LinkedList<ExecutionContext> newRuns = new LinkedList<>();
         for (ExecutionContext oldContext : contexts) {
             DataSet ds = oldContext.getDataSet();
-            if (!test && !checkJobStartability(ds)) continue;
+            if (!test && !canStartJobs(ds)) {
```
dankwart-de: I see why you did this, but now you have code duplication

vinjana: Could be, but in this situation the method is a simple checker and should not have any side effects. If we want to remove the single line code duplication (the method is called at two places), we could introduce another method of at least two lines. So here I think functional stringency over minor code duplication.

PR #187, opened 2018-01-24
Fix recursive config evaluation

vinjana:


vinjana:
BatchEuphoria code change required to get this running.

RoddyCore/src/de/dkfz/roddy/execution/io/ExecutionService.groovy
```diff
@@ -181,7 +180,7 @@ abstract class ExecutionService implements BEExecutionService {
         String cmdString
         if (!configurationDisallowsJobSubmission && !allJobsBlocked && !pidIsBlocked && !preventCalls && !isDummyCommand) {
             try {
-                cmdString = command.toString()
+                cmdString = command.toBashCommandString()
```
dankwart-de: I don't think it is appropriate to call it that way. Bash is an interpreter right?

vinjana: Indeed, it is. Maybe a service class to actually do this conversion from class fields to bash command line would be better.

dankwart-de: Ok, I'll approve it and will open an issue for this topic. Too big to do it now.

PR #191, opened 2018-01-26
Pass env property

vinjana:
Depends on https://github.com/eilslabs/RoddyToolLib/pull/16 and https://github.com/eilslabs/Roddy-Default-Plugin/pull/5. Please check them first.

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/Roddy.java
+++ b/RoddyCore/src/de/dkfz/roddy/Roddy.java
@@ -717,7 +717,7 @@ public class Roddy {
                         .setCreateDaemon(true)
                         .setStrictMode(false)
                         .setTrackUserJobsOnly(trackUserJobsOnly)
-                        .setPassEnvironment(Optional.of(false))
+                        .setPassEnvironment(true)
                         .setTrackOnlyStartedJobs(trackOnlyStartedJobs)
 //                        .setRequestMemoryIsEnabled(RoddyConversionHelperMethods.toBoolean(getApplicationProperty(RunMode.CLI, "requestMemoryIsEnabled", "true"), true))
 //                        .setRequestMemoryIsEnabled(RoddyConversionHelperMethods.toBoolean(getApplicationProperty(RunMode.CLI, "requestWalltimeIsEnabled", "true"), true))
--- a/RoddyCore/src/de/dkfz/roddy/Roddy.java
+++ b/RoddyCore/src/de/dkfz/roddy/Roddy.java
@@ -26,10 +26,7 @@ import de.dkfz.roddy.execution.jobs.BatchEuphoriaJobManager;
 import de.dkfz.roddy.execution.jobs.JobManagerOptions;
 import de.dkfz.roddy.execution.jobs.direct.synchronousexecution.DirectSynchronousExecutionJobManager;
 import de.dkfz.roddy.plugins.LibrariesFactory;
-import de.dkfz.roddy.tools.AppConfig;
-import de.dkfz.roddy.tools.LoggerWrapper;
-import de.dkfz.roddy.tools.RoddyConversionHelperMethods;
-import de.dkfz.roddy.tools.RoddyIOHelperMethods;
+import de.dkfz.roddy.tools.*;
 import groovy.transform.CompileStatic;
 
 import java.io.File;
@@ -62,7 +59,7 @@ public class Roddy {
     private static String SETTINGS_DIRECTORY_NAME = ".roddy";
 
     private static final LoggerWrapper logger = LoggerWrapper.getLogger(Roddy.class.getSimpleName());
-    private static AppConfig applicationProperties;
+    private static RoddyAppConfig applicationProperties = new RoddyAppConfig();
 
     private static RunMode runMode;
     private static boolean mainStarted = false;
@@ -547,8 +544,8 @@ public class Roddy {
      * TODO Does this work from Windows?
      * This uses the proxy vole library.
      */
-    private static void initializeProxySettings() {
-        boolean useProxyForInternetConnection = Boolean.parseBoolean(Roddy.getOrSetApplicationProperty(Constants.APP_PROPERTY_NET_USEPROXY, FALSE));
+    private static void initializeProxySettings() throws ConfigurationError {
+        boolean useProxyForInternetConnection = applicationProperties.getOrSetBooleanApplicationProperty(Constants.APP_PROPERTY_NET_USEPROXY, false);
         if (useProxyForInternetConnection) {
 
             System.setProperty("java.net.useSystemProxies", "true");
@@ -615,7 +612,7 @@ public class Roddy {
 
     @Deprecated
     private static void warnDeprecatedRoddyScriptVariable() {
-        String deprecatedDefaultScratchDir = Roddy.getOrSetApplicationProperty(ConfigurationConstants.CVALUE_DEFAULT_SCRATCH_DIR, "");
+        String deprecatedDefaultScratchDir = applicationProperties.getOrSetApplicationProperty(ConfigurationConstants.CVALUE_DEFAULT_SCRATCH_DIR, "");
         if (!deprecatedDefaultScratchDir.equals("")) {
             System.err.println("WARNING: You did set the deprecated " + ConfigurationConstants.CVALUE_DEFAULT_SCRATCH_DIR + " to '" +
                     deprecatedDefaultScratchDir + ".\nPlease use " + Constants.APP_PROPERTY_SCRATCH_BASE_DIRECTORY + ".");
@@ -628,7 +625,7 @@ public class Roddy {
      */
     private static void setScratchDirectory(RecursiveOverridableMapContainerForConfigurationValues configurationValues) {
         warnDeprecatedRoddyScriptVariable();
-        String scratchBaseDir = Roddy.getOrSetApplicationProperty(Constants.APP_PROPERTY_SCRATCH_BASE_DIRECTORY, "");
+        String scratchBaseDir = applicationProperties.getOrSetApplicationProperty(Constants.APP_PROPERTY_SCRATCH_BASE_DIRECTORY, "");
         if (scratchBaseDir.equals("")) {
             File propertyFilePath = null;
             try {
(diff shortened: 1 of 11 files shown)


RoddyCore/src/de/dkfz/roddy/Roddy.java
```diff
@@ -618,17 +618,17 @@ private static void setDefaultRoddyQueueVariable(RecursiveOverridableMapContaine
      * @return
      */
     private static void setScratchDirectory(RecursiveOverridableMapContainerForConfigurationValues configurationValues) {
-        String deprecatedDefaultScratchDir = Roddy.getApplicationProperty(ConfigurationConstants.CVALUE_DEFAULT_SCRATCH_DIR, "");
+        String deprecatedDefaultScratchDir = Roddy.getOrSetApplicationProperty(ConfigurationConstants.CVALUE_DEFAULT_SCRATCH_DIR, "");
         if (!deprecatedDefaultScratchDir.equals("")) {
```
dankwart-de: If this leads to a deprecation warning, you might also put it in a separate method and mark this as deprecated as well. This way, we'll also get a deprecation warning on compile time.

RoddyCore/src/de/dkfz/roddy/Roddy.java
```diff
@@ -717,13 +717,13 @@ public static void initializeJobManager(RoddyStartupModes mode)
                         .setCreateDaemon(true)
                         .setStrictMode(false)
                         .setTrackUserJobsOnly(trackUserJobsOnly)
-                        .setPassEnvironment(true)
+                        .setPassEnvironment(Roddy.getOrSetBooleanApplicationProperty(Constants.APP_PROPERTY_JOB_MANAGER_PASS_ENVIRONMENT, false))
                         .setTrackOnlyStartedJobs(trackOnlyStartedJobs)
-//                        .setRequestMemoryIsEnabled(RoddyConversionHelperMethods.toBoolean(getApplicationProperty(RunMode.CLI, "requestMemoryIsEnabled", "true"), true))
```
dankwart-de: Remove this

RoddyCore/src/de/dkfz/roddy/Roddy.java
```diff
@@ -794,34 +794,73 @@ public static void loadPropertiesFile() {
...
+        getOrSetApplicationProperty("configurationDirectories");
     }
 
     public static AppConfig getApplicationConfiguration() {
         return applicationProperties;
     }
 
-    public static String getApplicationProperty(String pName) {
-        return getApplicationProperty(pName, "");
+    public static String getOrSetApplicationProperty(String pName) {
```
dankwart-de: We might be able to get rid of the runmode prefix in the ini files:
- Parse over all values in the loaded ini and remove the CLI. prefix (or store a copy without the prefix).
- Get rid of the runmode part in all the method headers.
- Ignore prefixes other than CLI. when loading inis


Should make ini files clearer and still backward compatible.

RoddyCore/src/de/dkfz/roddy/Roddy.java
```diff
@@ -794,34 +794,73 @@ public static void loadPropertiesFile() {
...
     }
 
-    public static String getApplicationProperty(RunMode runMode, String pName, String defaultValue) {
-        return getApplicationProperty(runMode.name() + "." + pName, getApplicationProperty(pName, defaultValue));
+    public static String getOrSetApplicationProperty(RunMode runMode, String pName, String defaultValue) {
+        return getOrSetApplicationProperty(runMode.name() + "." + pName, getOrSetApplicationProperty(pName, defaultValue));
     }
 
-    public static String getApplicationProperty(String pName, String defaultValue) {
+    public static String getOrSetApplicationProperty(String pName, String defaultValue) {
```
dankwart-de: Do you think it makes sense to move the verification code below to the INI loader class?

RoddyCore/src/de/dkfz/roddy/Roddy.java
```diff
@@ -991,7 +1030,7 @@ public static String getUsedRoddyVersion() {
                 result = clc.getOptionValue(opt);
         }
         if (!RoddyConversionHelperMethods.isNullOrEmpty(result)) return result;
-        return getApplicationProperty("useRoddyVersion", LibrariesFactory.PLUGIN_VERSION_CURRENT);
+        return getOrSetApplicationProperty("useRoddyVersion", LibrariesFactory.PLUGIN_VERSION_CURRENT);
```
dankwart-de: Is there a constant for useRoddyVersion? If not, leave it like it is.

RoddyCore/src/de/dkfz/roddy/execution/io/ExecutionService.groovy
```diff
@@ -103,7 +103,7 @@ abstract class ExecutionService implements BEExecutionService {
         ClassLoader classLoader = LibrariesFactory.getGroovyClassLoader()
 
         RunMode runMode = Roddy.getRunMode()
-        String executionServiceClassID = Roddy.getApplicationProperty(runMode, Constants.APP_PROPERTY_EXECUTION_SERVICE_CLASS, SSHExecutionService.class.getName())
+        String executionServiceClassID = Roddy.getOrSetApplicationProperty(runMode, Constants.APP_PROPERTY_EXECUTION_SERVICE_CLASS, SSHExecutionService.class.getName())
```
dankwart-de: Seeing all these Roddy. accesses makes me whonder, if we can just initialize the execution services differently. This would lead to less complexity.

RoddyCore/src/de/dkfz/roddy/Roddy.java
vinjana: done

RoddyCore/src/de/dkfz/roddy/Roddy.java
vinjana: done

RoddyCore/src/de/dkfz/roddy/Roddy.java
vinjana: Yes, but thought I postpone this (scope and time). Now it's implemented anyway, although not without effort.

RoddyCore/src/de/dkfz/roddy/execution/io/ExecutionService.groovy
vinjana: Maybe again the `@Builder` annotation to generate a parameter object?

RoddyCore/src/de/dkfz/roddy/Roddy.java
```diff
@@ -547,8 +544,8 @@ public static final boolean initializeServices(RoddyStartupModes mode) {
      * TODO Does this work from Windows?
      * This uses the proxy vole library.
      */
-    private static void initializeProxySettings() {
-        boolean useProxyForInternetConnection = Boolean.parseBoolean(Roddy.getOrSetApplicationProperty(Constants.APP_PROPERTY_NET_USEPROXY, FALSE));
+    private static void initializeProxySettings() throws ConfigurationError {
```
dankwart-de: We could try to differentiate between ConfigurationErrors, which come at a later point of execution, and InitiliazationError, which would be here. What do you think?

RoddyCore/src/de/dkfz/roddy/Roddy.java
```diff
@@ -955,7 +881,7 @@ public static File getCompressedAnalysisToolsOverviewFile() {
         if (getCommandLineCall().isOptionSet(option))
```
dankwart-de: What's your oppinion about moving all the configuration related methods to the new RoddyAppConfig class?
Then Roddy would be much leaner and we'd also have an object which we could pass around instead of accessing the static Roddy singleton

RoddyCore/src/de/dkfz/roddy/client/cliclient/RoddyCLIClient.groovy
```diff
@@ -737,26 +737,30 @@ public class RoddyCLIClient {
     /**
      * Sets up Roddys property file from the command line.
      */
-    static void performCommandLineSetup() {
+    static void performCommandLineSetup() throws ConfigurationError {
```
dankwart-de: I'm not sure whether we should get rid of this method or leave it. Currently the AppConfig class cannot write ini files and the only thing where input is necessary is for the keyword based setup.

RoddyCore/src/de/dkfz/roddy/config/RoddyAppConfig.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    public String getOrSetApplicationProperty(RunMode runMode, String pName, String defaultValue) {
+        return getOrSetApplicationProperty(runMode.name() + "." + pName,
+                getOrSetApplicationProperty(pName, defaultValue));
+    }
+
+    public boolean getOrSetBooleanApplicationProperty(String pName, boolean defaultValue) throws ConfigurationError {
+        try {
+            return Boolean.parseBoolean(getOrSetApplicationProperty(pName, String.valueOf(defaultValue), "boolean"));
```
dankwart-de: Maybe introduce an enum for the type?

RoddyCore/src/de/dkfz/roddy/Roddy.java
vinjana: I agree. Configuration is not `Roddy`'s responsibility and a dedicated application configuration object would be more suited. Could be a step towards getting rid of the singleton.

RoddyCore/src/de/dkfz/roddy/client/cliclient/RoddyCLIClient.groovy
vinjana: What do you mean by "keyword-based setup"?

RoddyCore/src/de/dkfz/roddy/config/RoddyAppConfig.java
vinjana: Yes, I was also thinking about that. But then I thought we should first get a better overview about where such code is located, and then centralize this kind of code with a good conceptual idea in the background. So I left further refactoring for later.

RoddyCore/src/de/dkfz/roddy/client/cliclient/RoddyCLIClient.groovy
dankwart-de: If "password" is setup as auth method.

RoddyCore/src/de/dkfz/roddy/config/RoddyAppConfig.java
dankwart-de: Sure

PR #192, opened 2018-01-29
Fix cvbundle

vinjana:


dist/bin/current/helperScripts/compileRoddyPlugin.sh
```diff
@@ -84,6 +84,6 @@ then
     jar cf $outputDirectory/$libName.jar $classFiles
     cd $outputDirectory
     [[ -d ".svn" ]] && svn add $libName.jar buildversion.txt #2> /dev/null
-    [[ -d ".git" ]] && git add $libName.jar buildversion.txt `git st | grep Plugin.java | cut -d ":" -f 2` #2> /dev/null
+    [[ -d ".git" ]] && git add $libName.jar buildversion.txt `git stage | grep Plugin.java | cut -d ":" -f 2` #2> /dev/null
```
dankwart-de: I am not sure if we should leave this in the code.

vinjana: This is just a quick-fix because `st` is just defined in .gitconfig on some systems. Plugin and in parts also Roddy compilation and versioning has to be refactored and adapted anyway, but not now.

PR #194, opened 2018-01-29
Documentation update.

vinjana:


docs/installationGuide.rst
```diff
@@ -34,13 +36,13 @@ To install and run Roddy the following programs need to be installed on your com
...
-```bash
-curl -s "https://get.sdkman.io" | bash
-sdk install groovy 2.4.13
-sdk install java 8u151-oracle
-```
+.. code-block:: Bash
+
+    curl -s "https://get.sdkman.io" | bash
+    sdk install groovy 2.4.13
+    sdk install java 8u151-oracle
```
dankwart-de: As mentioned, OpenJDK is also good.

docs/usersGuide/walkthrough.rst
```diff
@@ -71,12 +71,10 @@ The ini files are explained in detail in :doc:`../config/applicationProperties`.
 
 .. code-block:: ini
 
-    [COMMON]
```
dankwart-de: Why did you remove this? I think it should still be explained.

vinjana: 1. Maybe this option is not so important and can be explained in the applicationProperties.rst.
2. It was not explained here anyway.

If this is supposed to be a kind of minimal walkthrough, this option/section may also be dropped. What do you think?

dankwart-de: Ah ok, I see it. It's ok then.

PR #195, opened 2018-01-29
Set baseEnvironmentScript variable via -env only if it is actually set.

vinjana:


RoddyCore/src/de/dkfz/roddy/execution/jobs/Job.groovy
```diff
@@ -521,8 +521,10 @@ class Job extends BEJob<BEJob, BEJobResult> {
         }
         this.parameters.put(DEBUG_WRAP_IN_SCRIPT, parameterObjectToString(DEBUG_WRAP_IN_SCRIPT, debugWrapInScript))
 
-        this.parameters.put(APP_PROPERTY_BASE_ENVIRONMENT_SCRIPT,
-                Roddy.applicationConfiguration.getOrSetApplicationProperty(APP_PROPERTY_BASE_ENVIRONMENT_SCRIPT, ""))
+        if (Roddy.applicationConfiguration.containsKey(APP_PROPERTY_BASE_ENVIRONMENT_SCRIPT)) {
```
dankwart-de: But that means, that there won't be a default value in the cfg right? Is that intended? What about else?

vinjana: Exactly. No default value. The interference with the login process is pretty hefty (/etc/profile will usually be sourced!). That should be the users arbitrary decision -- not ours.

dankwart-de: Ok, then I accept :D

PR #200, opened 2018-02-05
Removed usesAutoCheckpoint, added TOOL_ID to essential job parameters…

vinjana:
…, and check in getResourceSetFromConfiguration() for UNKNOWN tool id. Otherwise small refactorings.


docs/config/applicationProperties.rst
```diff
@@ -40,6 +40,7 @@ Let's have a brief look at it:
     #executionServiceClass=de.dkfz.roddy.execution.io.SSHExecutionService
     executionServiceHost=[YOURHOST]
     executionServiceAuth=keyfile
+    #executionServiceKeyfileLocation=[keyfile path]   # use $HOME/.ssh/id_rsa by default
```
dankwart-de: There is also the default app ini (and also xmls) in dist/bin/current/helper.../
you can also add things there.

PR #202, opened 2018-02-06
Fix invalid job status lines in jobStateLogFile

dankwart-de:
The tool id was missing in these lines.


dankwart-de:
In the wrapper it is always used with 4 fields. Wrapper code and Roddy code should match at this point but it didn't.

PR #213, opened 2018-03-06
Improve FileGroup usability

dankwart-de:
Implement the Iterable interface and thus allow operations like:
for in
.forEach
for file groups. The methods effectively use the iterator methods of the
contained file list.

Also allow the groovy methods:
+
+=
<<
on FileGroups


dankwart-de:
There are some code errors... ignore it.

PR #214, opened 2018-03-06
Improve FileGroup usability

dankwart-de:
Implement the Iterable interface and thus allow operations like:
for in
.forEach
for file groups. The methods effectively use the iterator methods of the
contained file list.

Also allow the groovy methods:
+
+=
<<
on FileGroups

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/knowledge/files/FileGroup.java
+++ b/RoddyCore/src/de/dkfz/roddy/knowledge/files/FileGroup.java
@@ -9,8 +9,11 @@ package de.dkfz.roddy.knowledge.files;
 import de.dkfz.roddy.config.ConfigurationError;
 import de.dkfz.roddy.core.ExecutionContext;
 
+import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


RoddyCore/src/de/dkfz/roddy/knowledge/files/FileGroup.java
```diff
@@ -32,6 +35,21 @@ public FileGroup(List<F> files) {
         }
     }
 
+    @Override
+    public Iterator<F> iterator() {
+        return filesInGroup.iterator();
+    }
+
+    @Override
+    public void forEach(Consumer<? super F> action) {
```
vinjana: Interesting that this one also has to be overridden despite the iterator() override.

dankwart-de: I actually don't know. I think this overrides the Java .forEach method whereas iterator is the one used by "for in" or Groovy .each

PR #231, opened 2018-03-25
Added up to Tuple10.

vinjana:
My biggest question concerns Tuple13 and the field naming.

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/knowledge/files/Tuple13.java
+++ b/RoddyCore/src/de/dkfz/roddy/knowledge/files/Tuple13.java
@@ -27,6 +27,7 @@ public class Tuple13<
         M extends FileObject
         > extends AbstractFileObjectTuple {
 
+    // TODO: @dankwart-de: Why has Tuple13 different names for the fields then e.g. Tuple2??
     public final A _a;
     public final B _b;
     public final C _c;
--- a/RoddyCore/src/de/dkfz/roddy/knowledge/files/Tuple7.java
+++ b/RoddyCore/src/de/dkfz/roddy/knowledge/files/Tuple7.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 eilslabs.
+ * Copyright (c) 2018 German Cancer Research Center (DKFZ).
  *
  * Distributed under the MIT License (license terms are at https://www.github.com/eilslabs/Roddy/LICENSE.txt).
  */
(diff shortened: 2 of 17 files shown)


RoddyCore/src/de/dkfz/roddy/knowledge/files/Tuple13.java
```diff
@@ -27,6 +27,7 @@
         M extends FileObject
         > extends AbstractFileObjectTuple {
 
+    // TODO: @dankwart-de: Why has Tuple13 different names for the fields then e.g. Tuple2??
```
dankwart-de: For that, you need to know, how I named the tuple fields in the first tuple implementations. I took the names from computer graphics vectors, namely x, y, z and w / r (or else). As you can see above, this mechanism gets confusing and misleading for higher order vectors and stops working at some point. That's why I started the tuple13 with _a

RoddyCore/src/de/dkfz/roddy/knowledge/files/Tuple8.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright (c) 2016 eilslabs.
```
dankwart-de: It is not eilslabs anymore. It is "The Roddy WMS developers" or DKFZ or so.

RoddyCore/src/de/dkfz/roddy/knowledge/files/Tuple9.java
```diff
@@ -0,0 +1,30 @@
...
+        R extends FileObject,
+        Q extends FileObject,
+        W extends FileObject,
+        S extends FileObject,
+        T extends FileObject,
+        U extends FileObject>
+        extends Tuple8<X, Y, Z, R, Q, W, S, T> {
+    private final U value8;
+
+    public Tuple9(X value0, Y value1, Z value2, R value3, Q value4, W value5, S value6, T value7, U value8) {
```
dankwart-de: As these are all new objects, I'd vote for a better naming scheme (like in tuple13). Maybe go from _a to ... _z / a ... z. Here the order is confusing.

RoddyCore/src/de/dkfz/roddy/knowledge/files/Tuple7.java
```diff
@@ -9,17 +9,19 @@
 /**
  * A file object tuple designed for (generic) job calls with 5 entries
```
vinjana: fix comment

PR #232, opened 2018-03-25
Optional build number increase

vinjana:


vinjana:
e.g. INCREASE_BUILD_NUMBER=false roddy.sh compileplugin XY

PR #246, opened 2018-04-20
Add checks for i/o directories and fail on errors

dankwart-de:
If both dirs are empty, abort.

If one of the dirs is set, also set its value for the other one.

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
+++ b/RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
@@ -494,13 +494,39 @@ class ProjectLoader {
         List<String> errors = []
 
         // Earliest check for valid input and output directories. If they are not accessible or writeable.
-        // Start with the input directory
-        errors += checkDirForReadabilityAndExecutability(analysis.getInputBaseDirectory(), "input")
-        errors += checkDirForReadabilityAndExecutability(analysis.getOutputBaseDirectory(), "output")
 
-        // Out dir needs to be writable
-        if (!FileSystemAccessProvider.instance.isWritable(analysis.getOutputBaseDirectory()))
-            errors << (String) "The output was not writeable at path ${analysis.getOutputBaseDirectory()}."
+        // Check if the configuration values are both set, print out warning if one was set empty.
+        // If one is empty, fill with the other value, if both are empty, abort!
+        // Checks are done before the readability tests because we need to check for the raw configuration values. The next
+        // check will already use translated value
+
+        String valueInDir = analysis.configuration.configurationValues.get(ConfigurationConstants.CFG_INPUT_BASE_DIRECTORY, "").value
+        String valueOutDir = analysis.configuration.configurationValues.get(ConfigurationConstants.CFG_OUTPUT_BASE_DIRECTORY, "").value
+        
+        if (!valueInDir && !valueOutDir) {
+            errors << "Both the input and output base directories are not set. You must set at least inputBaseDirectory or outputBaseDirectory."
+
+        } else {
+
+            // Fill variable, if it is missing. Log a warning.
+            if(!valueInDir) {
+                logger.always("The input base directory is not set, take the path of the output base directory for it.")
+                analysis.configuration.configurationValues.add(new ConfigurationValue(ConfigurationConstants.CFG_INPUT_BASE_DIRECTORY, valueOutDir, "path"))
+            }
+
+            if(!valueOutDir) {
+                logger.always("The output base directory is not set, take the path of the input base directory for it.")
+                analysis.configuration.configurationValues.add(new ConfigurationValue(ConfigurationConstants.CFG_OUTPUT_BASE_DIRECTORY, valueInDir, "path"))
+            }
+
+            // Now start with the input directory
+            errors += checkDirForReadabilityAndExecutability(analysis.getInputBaseDirectory(), "input")
+            errors += checkDirForReadabilityAndExecutability(analysis.getOutputBaseDirectory(), "output")
+
+            // Out dir needs to be writable
+            if (!FileSystemAccessProvider.instance.isWritable(analysis.getOutputBaseDirectory()))
+                errors << (String) "The output was not writeable at path ${analysis.getOutputBaseDirectory()}."
+        }
 
         if (!errors)
             return


RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -494,13 +494,39 @@ class ProjectLoader {
...
+        String valueOutDir = analysis.configuration.configurationValues.get(ConfigurationConstants.CFG_OUTPUT_BASE_DIRECTORY, "").value
+        
+        if (!valueInDir && !valueOutDir) {
+            errors << "Both the input and output base directories are not set. You must set at least inputBaseDirectory or outputBaseDirectory."
+
+        } else {
+
+            // Fill variable, if it is missing. Log a warning.
+            if(!valueInDir) {
+                logger.always("The input base directory is not set, take the path of the output base directory for it.")
```
vinjana: The message sounds like a request, but actually you are doing the "requested" change already in the code. Therefore 

"The input base directory is not set. Taking the path of the output base directory instead."

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -494,13 +494,39 @@ class ProjectLoader {
...
+                analysis.configuration.configurationValues.add(new ConfigurationValue(ConfigurationConstants.CFG_INPUT_BASE_DIRECTORY, valueOutDir, "path"))
+            }
+
+            if(!valueOutDir) {
+                logger.always("The output base directory is not set, take the path of the input base directory for it.")
```
vinjana: dito

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -494,13 +494,39 @@ class ProjectLoader {
...
 
         // Earliest check for valid input and output directories. If they are not accessible or writeable.
-        // Start with the input directory
-        errors += checkDirForReadabilityAndExecutability(analysis.getInputBaseDirectory(), "input")
-        errors += checkDirForReadabilityAndExecutability(analysis.getOutputBaseDirectory(), "output")
 
-        // Out dir needs to be writable
-        if (!FileSystemAccessProvider.instance.isWritable(analysis.getOutputBaseDirectory()))
-            errors << (String) "The output was not writeable at path ${analysis.getOutputBaseDirectory()}."
+        // Check if the configuration values are both set, print out warning if one was set empty.
```
vinjana: Comments are not necessary. Remove. If you think the code does not tell this simple story, rather improve the code.

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -494,13 +494,39 @@ class ProjectLoader {
...
+        // If one is empty, fill with the other value, if both are empty, abort!
+        // Checks are done before the readability tests because we need to check for the raw configuration values. The next
+        // check will already use translated value
+
+        String valueInDir = analysis.configuration.configurationValues.get(ConfigurationConstants.CFG_INPUT_BASE_DIRECTORY, "").value
```
vinjana: Why "value..."? I see that the evaluation is necessary, for the subsequent fFS checks, but see my comment further down.

Actually, we seem to have some work confusion here. We have configuration values that additionally can be evaluated. The method `getValue()` is responsible for evaluating an value. *Urgh*

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -494,13 +494,39 @@ class ProjectLoader {
...
+        
+        if (!valueInDir && !valueOutDir) {
+            errors << "Both the input and output base directories are not set. You must set at least inputBaseDirectory or outputBaseDirectory."
+
+        } else {
+
+            // Fill variable, if it is missing. Log a warning.
+            if(!valueInDir) {
+                logger.always("The input base directory is not set, take the path of the output base directory for it.")
+                analysis.configuration.configurationValues.add(new ConfigurationValue(ConfigurationConstants.CFG_INPUT_BASE_DIRECTORY, valueOutDir, "path"))
```
vinjana: Why do you put in the evaluated configuration value here? Wouldn't it be wiser to postpone the evaluation to the use site?

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -494,13 +494,36 @@ class ProjectLoader {
...
+        } else {
+
+            // Fill variable, if it is missing. Log a warning.
+            if(!valueInDir) {
+                logger.always("The input base directory is not set. Taking the path of the output base directory instead.")
+                analysis.configuration.configurationValues.add(new ConfigurationValue(ConfigurationConstants.CFG_INPUT_BASE_DIRECTORY, valueOutDir, "path"))
+            }
+
+            if(!valueOutDir) {
+                logger.always("The output base directory is not set. Taking the path of the output base directory instead.")
```
vinjana: "Taking the value of the *input* base directory instead."

RoddyCore/src/de/dkfz/roddy/core/ProjectLoader.groovy
```diff
@@ -494,13 +494,36 @@ class ProjectLoader {
...
-        // Out dir needs to be writable
-        if (!FileSystemAccessProvider.instance.isWritable(analysis.getOutputBaseDirectory()))
-            errors << (String) "The output was not writeable at path ${analysis.getOutputBaseDirectory()}."
+        String valueInDir = analysis.configuration.configurationValues.get(ConfigurationConstants.CFG_INPUT_BASE_DIRECTORY, "").value
+        String valueOutDir = analysis.configuration.configurationValues.get(ConfigurationConstants.CFG_OUTPUT_BASE_DIRECTORY, "").value
+        
+        if (!valueInDir && !valueOutDir) {
+            errors << "Both the input and output base directories are not set. You must set at least inputBaseDirectory or outputBaseDirectory."
+
+        } else {
```
vinjana: To have less nesting, why not `} else if (...) {`?

dankwart-de: Sorry, don't get it. Why introduce another if?

PR #256, opened 2018-05-29
Improved unknown libdir during compile

vinjana:


RoddyCore/test/src/de/dkfz/roddy/config/converters/BashConverterTest.groovy
```diff
@@ -17,14 +19,18 @@ import de.dkfz.roddy.execution.io.NoNoExecutionService
...
+import org.junit.rules.TemporaryFolder
 
 /**
  * Created by heinold on 30.06.16.
  */
 @groovy.transform.CompileStatic
 class BashConverterTest {
 
+    @Rule
+    public TemporaryFolder testFolder = new TemporaryFolder()
```
dankwart-de: Even a TemporaryFolder is not necessarily removed. If you want to make sure, that it is removed, create a TearDown method.

PR #262, opened 2018-06-20
Improve cohort code

dankwart-de:
Code is already running and in use for two colleagues.


dankwart-de:
Merged due to unability to review in time.

PR #264, opened 2018-07-03
Fix missing synchronized get filename

dankwart-de:
See commit message


vinjana:
What is this PR for? Could you explain a bit? When does the error occur?

vinjana:
Also the CI fails.

Issue #79, opened 2017-07-26
Are cvalue recursions properly detected?


dankwart-de:
Like e.g.
cvalue name="A_BINARY" value="$A_BINARY" ?

dankwart-de:
Actually it somehow is. But it throws a RuntimeException which is catched in Roddys main method. 
Unfortunately, we use the toString method and the ConfigurationValue class is not a Groovy class. Therefore we cannot change the RuntimeException to e.g. a ConfigurationException.
Unfortunately as well, even if we detect the recursion, remove the throw and break properly with a message to the configuration, this message is just ignored.

Don't know, maybe we try to append the message to the execution service? Might be the proper place. Change the exception type to ConfigurationError would be right, but not trivial.

vinjana:
Seems like we should first get away from the `catch ([Runtime]Exception e)` in the code. Then we solve this problem.

vinjana:
This should not just be checked. The checks need to be implemented directly as unit tests.

dankwart-de:
Working on it.
Unit tests are there and working.
Try to output it better and get rid of ugly stacktraces.

dankwart-de:
See #278

dankwart-de:
Currently on hold...

PR #278, opened 2018-07-13
Improve cyclic dependency detection and err output

dankwart-de:
Removed an instruction from RoddyCLIClient.outputtestrunresult. This
line converted the configuration to what? So I got rid of it. The line
showed exceptions with cyclic dependencies.

Configuration.getID() will return "Unnamed configuration", if no
preloadedConfiguration is set.

Moved ConfigurationValue.java => ConfigurationValue.groovy
- Escape all kinds of $ to \$ otherwise replacements won't work.
- Instead of a RuntimeException, throw a ConfigurationError now when it
  comes to cyclic dependencies

In Aalysis class, upon run, catch ConfigurationError as more some type
of a known error. The developer needs to add ConfigurationError
instances to the configuration so they get printed!

Let ExecutionContext.execute throw a ConfigurationError

SimpleCohortWorkflow now contains a small cyclic dependency.


RoddyCore/src/de/dkfz/roddy/config/Configuration.java
```diff
@@ -201,7 +201,10 @@ public String getName() {
      */
     @Override
     public String getID() {
-        return preloadedConfiguration.id;
+        if(preloadedConfiguration != null)
+            return preloadedConfiguration.id;
+        else
+            return "'Unnamed Configuration'";
```
vinjana: Why are there single ticks inside the double ticks?

RoddyCore/src/de/dkfz/roddy/config/ConfigurationValue.groovy
```diff
@@ -119,7 +121,7 @@ private String checkAndCorrectPathThrow(String temp) throws ConfigurationError {
         String curUserPath = (new File("")).getAbsolutePath();
         String applicationDirectory = Roddy.getApplicationDirectory().getAbsolutePath();
         //TODO Make something like a blacklist. This is not properly handled now. Initially this was done because Java sometimes puts something in front of the file paths.
-        if (value.startsWith("${") || value.startsWith("$") || value.startsWith("~") || !value.startsWith("/")) {
+        if (value.startsWith('${') || value.startsWith('$') || value.startsWith("~") || !value.startsWith("/")) {
```
vinjana: Now that you touched that code, would it not make sense generalizing it and moving it into e.g. a dedicated class. From there we could evolve it into a true templating system, or use an existing templating system. We use similar code also for the variables in the filename patterns. Also, there are plenty of repetitions here -- '${', etc.

RoddyCore/src/de/dkfz/roddy/core/Analysis.java
```diff
@@ -448,6 +451,9 @@ protected void executeRun(ExecutionContext context, boolean preventLoggingOnQuer
                     }
                 }
             }
+        } catch (ConfigurationError ce) {
```
vinjana: What? Actually, I also touched that code and more.

RoddyCore/src/de/dkfz/roddy/config/Configuration.java
dankwart-de: For readability.

dankwart-de:
This PR will be blocked. vinjana has a local branch with a lot changes at the same place! Leave for now.
Related to #79

RoddyCore/src/de/dkfz/roddy/config/Configuration.java
vinjana: They don't belong there. You need to add them to the place, where the string is printed.

RoddyCore/src/de/dkfz/roddy/config/ConfigurationValue.groovy
dankwart-de: It would. But not for now, that'll be a bigger discussion right?

vinjana: Right.

dankwart-de:
Discussed with vinjana, will merge

Issue #279, opened 2018-07-13
Testrerun does not show all job parameters.


dankwart-de:
Testrun shows all parameters which are passed to a job during a call. However, testrerun does not.

PR #280, opened 2018-07-13
Fix three bugs

dankwart-de:
testrerun was not showing all details for a job. This was fixed by
adding allFinalParameters to the Job class.

Fix a test in ExecutionService.
validSubmissionCommandOutputWithRandomJobId did not have switches for
DirectCommand and SLURM. SLURM was not fixed (BE can't do it yet),
DirectCommand was.

For testrerun (like for testrun), the Direct Job Manager should be used,
however the target Job Manager was used. Change the startup mode scope
from SCOPE_FULL_WITH_JOBMANAGER to SCOPE_FULL for testrerun.

See #279


RoddyCore/src/de/dkfz/roddy/execution/io/fs/FileSystemAccessProvider.groovy
```diff
@@ -68,6 +68,11 @@ public class FileSystemAccessProvider {
      */
     protected File _userHome;
 
+    /**
+     * The cached "remote" umask
+     */
+    protected Integer _umask
```
vinjana: Then you call it `_remote_umask` or `_default_remote_umask` (see below)?

RoddyCore/src/de/dkfz/roddy/execution/io/fs/FileSystemAccessProvider.groovy
```diff
@@ -769,8 +774,11 @@ public class FileSystemAccessProvider {
     }
 
     int getDefaultUserMask() {
```
vinjana: Here it is a "default" umask, but the variable is calles `_umask`. Make it consistent -- either with or without "default", whatever fits better.

RoddyCore/src/de/dkfz/roddy/execution/jobs/Job.groovy
```diff
@@ -60,6 +60,12 @@ class Job extends BEJob<BEJob, BEJobResult> {
      */
     protected final Map<String, Object> allRawInputParameters
 
+    /**
+     * Keeps a list of all parameters after conversion and before removing them from the final parameter list
+     * (See keepOnlyEssentialParameters)
+     */
+    final Map<String, String> allFinalParameters = [:]
```
vinjana: What other purpose does this variable have, besides providing a parameter list for presentation to the user? Maybe it should be renamed to `reportedParameters`?

`allFinalParameters` is probably even not correct, because one could argue, that even those parameters in the _parameter file are final and should belong to the "all" set.

RoddyCore/src/de/dkfz/roddy/execution/io/fs/FileSystemAccessProvider.groovy
dankwart-de: I'll put the remote in braces (remote). It can but does not need to be remote. _default_umask absolutely makes sense.

RoddyCore/src/de/dkfz/roddy/execution/io/fs/FileSystemAccessProvider.groovy
dankwart-de: see above. Will rename it.

RoddyCore/src/de/dkfz/roddy/execution/jobs/Job.groovy
dankwart-de: Sound reasonable, will rename it.

PR #301, opened 2018-08-27
Fix printidlessruntimeconfig

vinjana:
* Moved IO-dir checks up to allow optional checking depending on startup mode
* Backport 3.0 branch to support ContextResource during tests


vinjana:
Integration tests succeeded.

PR #309, opened 2018-11-19
Fix for Configuration value replacement

dankwart-de:
ConfigurationValue:
- Fix the constructor for "double" values
- Add the replaceObsoleteVariableIdentifiers and call it from the
  constructor. The method replaces several older malformed values ($VAL)
  to their proper form (${VAL}).
- Remove the methods:
    * replaceString
    * toFile(), toFile(Map<String, String> replacements)
    * replaceConfigurationBasedValues
    * toString(List<String>)
- Consolidate and cleanup the methods toFile(Analysis, DataSet) and
  toFile(Analysis). The methods now use the normal variable replacement
  feature of Roddy instead of the removed replace method
- Make the field variableDetection static final
- Make toEvaluatedValue final and extend its argument list

Add a on-null-check to DefaultValidator.checkDollarSignUsage

Analysis:
- Clean up and groovyfy the class
- Change the getConfiguration method so that it will return a
  configuration which knows values like USERNAME and USERGROUP

DataSet now contains a getConfiguration method, which knows the
different variants of DATASET. For internal use only!

Add tests for variable replacement to ConfigurationValueSpec

Add ExecutionService and FileSystemAccessProvider init code to some
test classes

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/Constants.java
+++ b/RoddyCore/src/de/dkfz/roddy/Constants.java
@@ -1,7 +1,7 @@
 /*
- * Copyright (c) 2016 eilslabs.
+ * Copyright (c) 2016 German Cancer Research Center (Deutsches Krebsforschungszentrum, DKFZ).
  *
- * Distributed under the MIT License (license terms are at https://www.github.com/eilslabs/Roddy/LICENSE.txt).
+ * Distributed under the MIT License (license terms are at https://www.github.com/TheRoddyWMS/Roddy/LICENSE.txt).
  */
 
 package de.dkfz.roddy;
@@ -41,6 +41,7 @@ public class Constants {
     public static final String APP_PROPERTY_SCRATCH_BASE_DIRECTORY = "scratchBaseDirectory";
     public static final String APP_PROPERTY_JOB_MANAGER_PASS_ENVIRONMENT = "jobManagerPassEnvironment";
     public static final String APP_PROPERTY_JOB_MANAGER_HOLDJOBS_ON_SUBMISSION = "jobManagerHoldJobsOnSubmission";
+    public static final String APP_PROPERTY_JOB_MANAGER_UPDATE_INTERVAL = "jobManagerUpdateInterval";
     public static final String APP_PROPERTY_BASE_ENVIRONMENT_SCRIPT = "baseEnvironmentScript";
 
     /////////////////////////
@@ -50,8 +51,6 @@ public class Constants {
     public static final String ERR_MSG_ONLY_ONE_JOB_ALLOWED = "A job object is not allowed to run several times.";
     public static final String ERR_MSG_WRONG_PARAMETER_COUNT = "You did not provide proper parameters, args.length = ";
     public static final String ERR_MSG_NO_APPLICATION_PROPERTY_FILE = "Configuration does not exist. Cannot start application.";
-
-//    public static final String APP_EXITCODE_
 
     /////////////////////////
     // Environment settings
--- a/RoddyCore/src/de/dkfz/roddy/Constants.java
+++ b/RoddyCore/src/de/dkfz/roddy/Constants.java
@@ -34,7 +34,6 @@ public class Constants {
     public static final String APP_PROPERTY_CONFIGURATION_DIRECTORIES = "configurationDirectories";
     public static final String APP_PROPERTY_PLUGIN_DIRECTORIES = "pluginDirectories";
     public static final String APP_PROPERTIES_FILENAME = "applicationProperties.ini";
-    public static final String APP_PROPERTY_APPLICATION_DEBUG_TAG_NOJOBSUBMISSION = "NOJOBSUBMISSION";
     public static final String APP_PROPERTY_NET_USEPROXY = "netUseProxy";
     public static final String APP_PROPERTY_NET_PROXY_ADDRESS = "netProxyAddress";
     public static final String APP_PROPERTY_NET_PROXY_USR = "netProxyUser";
@@ -50,20 +49,12 @@ public class Constants {
 
     public static final String ERR_MSG_ONLY_ONE_JOB_ALLOWED = "A job object is not allowed to run several times.";
     public static final String ERR_MSG_WRONG_PARAMETER_COUNT = "You did not provide proper parameters, args.length = ";
-    public static final String ERR_MSG_NO_APPLICATION_PROPERTY_FILE = "Configuration does not exist. Cannot start application.";
 
     /////////////////////////
     // Environment settings
     /////////////////////////
 
-    public static final String ENV_LINESEPARATOR = System.getProperty("line.separator");
-
-    /////////////////////////
-    // Roddy tools
-    /////////////////////////
-
-    public static final String TOOLID_CREATE_LOCKFILES = "createLockFiles";
-    public static final String TOOLID_STREAM_BUFFER = "streamBuffer";
+    public static final String ENV_LINESEPARATOR = SystemProperties.getLineSeparator();
 
     /////////////////////////
     // Other constants
@@ -72,6 +63,9 @@ public class Constants {
     public static final String UNKNOWN_USER = "UNKNOWN";
     public static final String UNKNOWN = "UNKNOWN";
     public static final String NO_VALUE = "<NO_VALUE>";
+    public static final String USERNAME = "USERNAME";
+    public static final String USERGROUP = "USERGROUP";
+    public static final String USERHOME = "USERHOME";
     public static final String DEFAULT = "default";
 
     public static final String RODDY_PARENT_JOBS = "RODDY_PARENT_JOBS";
@@ -81,8 +75,11 @@ public class Constants {
     public static final String PARAMETER_FILE_SUFFIX = ".parameters";
     public static final String PROJECT_NAME = "projectName";
     public static final String DATASET = "dataSet";
+    public static final String DATASET_HR = "dataset";
     public static final String DATASET_CAP = "DATASET";
+    @Deprecated
     public static final String PID_CAP = "PID";
+    @Deprecated
     public static final String PID = "pid";
 
     /////////////////////////
--- a/RoddyCore/src/de/dkfz/roddy/Roddy.java
+++ b/RoddyCore/src/de/dkfz/roddy/Roddy.java
@@ -226,7 +226,7 @@ public class Roddy {
             // we check for this particular exception by using its simple class name!
             if (!e.getClass().getName().endsWith("SystemExitException"))
                 e.printStackTrace();
-            exit(ExitReasons.groovyServError);
+            exitWithMessage(ExitReasons.groovyServError);
         }
     }
 
@@ -429,7 +429,7 @@ public class Roddy {
                 }
             }
         } catch (RuntimeException e) {
-            exit(ExitReasons.unparseableStartupOptions);
+            exitWithMessage(ExitReasons.unparseableStartupOptions);
         }
     }
 
@@ -439,7 +439,7 @@ public class Roddy {
             toggleIni = new File(commandLineCall.getOptionValue(RoddyStartupOptions.usefeaturetoggleconfig));
             logger.postSometimesInfo("Trying to load alternative feature toggles file: " + toggleIni);
             if (toggleIni == null || !toggleIni.exists()) {
-                exit(ExitReasons.unknownFeatureToggleFile);
+                exitWithMessage(ExitReasons.unknownFeatureToggleFile);
             }
 
         } else {
@@ -779,7 +779,7 @@ public class Roddy {
                 }
             }
         } catch (Exception ex) {
-            exit(ExitReasons.waitForJobsFailedWithAnUnknownError);
+            exitWithMessage(ExitReasons.waitForJobsFailedWithAnUnknownError);
         }
         return 0;
     }
@@ -795,7 +795,7 @@ public class Roddy {
         System.exit(ecode <= 255 ? ecode : ExitReasons.wrongExitCodeUsed.getCode()); //Exit codes should be in range from 0 .. 255
     }
 
-    public static void exit(ExitReasons reason) {
+    public static void exitWithMessage(ExitReasons reason) {
         logger.severe(reason.getMessage());
         exit(reason.getCode());
     }
--- a/RoddyCore/src/de/dkfz/roddy/Roddy.java
+++ b/RoddyCore/src/de/dkfz/roddy/Roddy.java
@@ -1,7 +1,7 @@
 /*
- * Copyright (c) 2016 eilslabs.
+ * Copyright (c) 2016 German Cancer Research Center (Deutsches Krebsforschungszentrum, DKFZ).
  *
- * Distributed under the MIT License (license terms are at https://www.github.com/eilslabs/Roddy/LICENSE.txt).
+ * Distributed under the MIT License (license terms are at https://www.github.com/TheRoddyWMS/Roddy/LICENSE.txt).
  */
 
 package de.dkfz.roddy;
@@ -26,7 +26,10 @@ import de.dkfz.roddy.execution.jobs.BatchEuphoriaJobManager;
 import de.dkfz.roddy.execution.jobs.JobManagerOptions;
 import de.dkfz.roddy.execution.jobs.direct.synchronousexecution.DirectSynchronousExecutionJobManager;
 import de.dkfz.roddy.plugins.LibrariesFactory;
-import de.dkfz.roddy.tools.*;
+import de.dkfz.roddy.tools.AppConfig;
+import de.dkfz.roddy.tools.LoggerWrapper;
+import de.dkfz.roddy.tools.RoddyConversionHelperMethods;
+import de.dkfz.roddy.tools.RoddyIOHelperMethods;
 import groovy.transform.CompileStatic;
 
 import java.io.File;
@@ -37,6 +40,7 @@ import java.net.InetSocketAddress;
 import java.net.ProxySelector;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.time.Duration;
 import java.util.*;
 
 import static de.dkfz.roddy.RunMode.CLI;
@@ -201,7 +205,7 @@ public class Roddy {
             if (list == null || list.length != 1) {
                 System.out.println("You must call roddy from the right location! (Base roddy folder where roddy.sh resides). You used: '" +
                         applicationDirectory + "'");
-                System.exit(255);
+                System.exit(ExitReasons.wrongStartupLocation.getCode());
             }
 
             if (args.length == 0) {
@@ -222,7 +226,7 @@ public class Roddy {
             // we check for this particular exception by using its simple class name!
             if (!e.getClass().getName().endsWith("SystemExitException"))
                 e.printStackTrace();
-            exit(1);
+            exit(ExitReasons.groovyServError);
         }
     }
 
@@ -248,7 +252,7 @@ public class Roddy {
         CommandLineCall clc = new CommandLineCall(list);
         commandLineCall = clc;
         if (clc.isMalformed())
-            exit(1);
+            exit(ExitReasons.malformedCommandLine.getCode());
 
         // Initialize the logger with an initial setup. At this point we don't know about things like the logger settings
         // or the used app ini file. However, all the following methods rely on an existing valid logger setup.
@@ -258,7 +262,7 @@ public class Roddy {
 
         time("initial checks");
         if (!roddyExecutionRequirementsFulfilled())
-            exit(1);
+            exit(ExitReasons.unfulfilledRequirements.getCode());
 
         time("ftoggleini");
         initializeFeatureToggles();
@@ -281,7 +285,7 @@ public class Roddy {
         }
 
         if (!jobExecutionRequirementsFulfilled())
-            exit(1);
+            exit(ExitReasons.unfulfilledRequirements.getCode());
 
         time("initserv");
         parseRoddyStartupModeAndRun(clc);
@@ -425,8 +429,7 @@ public class Roddy {
                 }
             }
         } catch (RuntimeException e) {
-            logger.severe("Parsing startup options failed.");
-            exit(1);
+            exit(ExitReasons.unparseableStartupOptions);
         }
     }
 
@@ -436,8 +439,7 @@ public class Roddy {
             toggleIni = new File(commandLineCall.getOptionValue(RoddyStartupOptions.usefeaturetoggleconfig));
             logger.postSometimesInfo("Trying to load alternative feature toggles file: " + toggleIni);
             if (toggleIni == null || !toggleIni.exists()) {
-                logger.postAlwaysInfo("Cannot find requested toggle file.");
-                exit(1);
+                exit(ExitReasons.unknownFeatureToggleFile);
             }
 
         } else {
@@ -465,7 +467,7 @@ public class Roddy {
             String toggleNames = RoddyIOHelperMethods.joinArray(AvailableFeatureToggles.values(), "\n\t");
             logger.severe("Available toggle values are:\n\t" + toggleNames);
             if (isStrictModeEnabled())
-                exit(1);
+                exit(ExitReasons.unknownFeatureToggle.getCode());
         }
     }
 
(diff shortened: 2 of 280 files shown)


dankwart-de:
Note, that I compared the created configuration of this version and a previous Roddy version with test data and the Indel workflow.

RoddyCore/src/de/dkfz/roddy/config/ConfigurationValue.groovy
```diff
@@ -114,7 +115,7 @@ class ConfigurationValue implements RecursiveOverridableMapContainer.Identifiabl
 
     ConfigurationValue(Configuration config, String id, String value, String type, String description, List<String> tags) {
         this.id = id
-        this.value = value
+        this.value = value != null ? replaceObsoleteVariableIdentifiers(value) : null
```
vinjana: Would "deprecated" not be more fitting that "obsolete"?

RoddyCore/src/de/dkfz/roddy/config/ConfigurationValue.groovy
```diff
@@ -244,41 +255,12 @@ class ConfigurationValue implements RecursiveOverridableMapContainer.Identifiabl
      * @param dataSet
      * @return
      */
-    File toFile(Analysis analysis, DataSet dataSet) throws ConfigurationError {
-        File f = toFile(analysis)
+    File toFile(Analysis analysis, DataSet dataSet = null) throws ConfigurationError {
```
vinjana: The evaluation order here may be important and should maybe get documented.
 1. configuration itself (Not clear here from the code, what the first configuration is. Could be an analysis or context configuration as well, or not?)
 2. analysis-configuration
 3. dataset-configuration

RoddyCore/src/de/dkfz/roddy/config/validation/DefaultValidator.groovy
```diff
@@ -70,6 +70,9 @@ class DefaultValidator extends ConfigurationValueValidator {
      * API Level 3.4+
      */
     boolean checkDollarSignUsage(ConfigurationValue configurationValue) {
```
vinjana: My interpretation of "check" is positive/affirmative, i.e. if I check that something is used, it is present. The semantics of the method here are, however, negative -- true == **no** dollar sign.

Rename to e.g. "containsNoDollarSign" or so.

RoddyCore/src/de/dkfz/roddy/core/Analysis.groovy
```diff
@@ -32,142 +32,145 @@ import static de.dkfz.roddy.tools.RoddyIOHelperMethods.getStackTraceAsString
...
     }
 
-    public String getUsername() {
+    String getUsername() {
         try {
-            return FileSystemAccessProvider.getInstance().callWhoAmI();
+            return FileSystemAccessProvider.getInstance().callWhoAmI()
         } catch (Exception e) {
-            return "UNKNOWN";
+            return "UNKNOWN"
```
vinjana: BTW: Did this ever throw an exception? And if it does, is this not rather a real exception and should not be ignored like here?

RoddyCore/src/de/dkfz/roddy/core/Analysis.groovy
```diff
@@ -32,142 +32,145 @@ import static de.dkfz.roddy.tools.RoddyIOHelperMethods.getStackTraceAsString
...
-    public AnalysisConfiguration getConfiguration() {
-        if (_contextConfiguration == null)
-            _contextConfiguration = new ContextConfiguration((AnalysisConfiguration) this.configuration, (ProjectConfiguration) this.project.getConfiguration());
-        return _contextConfiguration;
-//        return configuration;
+    AnalysisConfiguration getConfiguration() {
+        if (_analysisConfiguration == null) {
+            _analysisConfiguration = new ContextConfiguration((AnalysisConfiguration) this.configuration, (ProjectConfiguration) this.project.getConfiguration())
+            [
+                    'USERNAME' : getUsername(),
```
vinjana: This is a repetition of the identifiers "USERNAME" etc. Better put them into consants.

RoddyCore/src/de/dkfz/roddy/core/Analysis.groovy
```diff
@@ -178,26 +181,26 @@ public class Analysis {
...
+        List<ExecutionContext> runs = new LinkedList<>()
 
-        long creationCheckPoint = System.nanoTime();
+        long creationCheckPoint = System.nanoTime()
 
         for (DataSet ds : selectedDatasets) {
             if (level.allowedToSubmitJobs && !canStartJobs(ds)) {
-                logger.postAlwaysInfo("The " + Constants.PID + " " + ds.getId() + " is still running and will be skipped for the process.");
-                continue;
+                logger.postAlwaysInfo("The " + Constants.PID + " " + ds.getId() + " is still running and will be skipped for the process.")
```
vinjana: Could be changed to `Constants.DATASET`. We wanted to get away from the PID concept, right?

RoddyCore/src/de/dkfz/roddy/core/Analysis.groovy
```diff
@@ -428,59 +431,59 @@ public class Analysis {
...
-                        boolean ignoreFileChecks = Roddy.isOptionSet(RoddyStartupOptions.disablestrictfilechecks);
-                        execute &= ignoreFileChecks || (invalidPreparedFiles.size() == 0 && copiedAnalysisToolsAreExecutable);
+                        List<String> invalidPreparedFiles = ExecutionService.getInstance().checkForInaccessiblePreparedFiles(context)
+                        boolean copiedAnalysisToolsAreExecutable = ExecutionService.getInstance().checkCopiedAnalysisTools(context)
+                        boolean ignoreFileChecks = Roddy.isOptionSet(RoddyStartupOptions.disablestrictfilechecks)
+                        execute &= ignoreFileChecks || (invalidPreparedFiles.size() == 0 && copiedAnalysisToolsAreExecutable)
                         if (!execute) {
-                            StringBuilder message = new StringBuilder("There were errors after preparing the workflow run for dataset " + datasetID);
+                            StringBuilder message = new StringBuilder("There were errors after preparing the workflow run for dataset " + datasetID)
```
vinjana: See previous comment. In principle just writing "dataset" may be sufficient in a message (compared to `Constants.DATASET`).

RoddyCore/src/de/dkfz/roddy/core/RuntimeService.groovy
```diff
@@ -471,7 +472,7 @@ class RuntimeService {
     }
 
     File getOutputAnalysisBaseDirectory(ExecutionContext context) {
-        return getOutputAnalysisBaseDirectory(context.dataSet, context.analysis)
+        return getOutputAnalysisBaseCV(context.analysis).toFile(context)
```
vinjana: `getOutputAnalysisBaseDirectoryCV(Analysis)` would be better, right?

RoddyCore/test/src/de/dkfz/roddy/config/ConfigurationValueSpec.groovy
```diff
@@ -24,4 +44,32 @@ class ConfigurationValueSpec extends Specification {
...
     }
+
+    def "test variable replacement for toFile()"() {
+        when:
+        ExecutionContext context = contextResource.createSimpleContext(ConfigurationValueSpec.class)
+        def values = context.configurationValues
+        values << new ConfigurationValue(context.configuration, "b", 'abc')
+        values << new ConfigurationValue(context.configuration, "c", 'def')
+        values << new ConfigurationValue(context.configuration, "a", '/tmp/${projectName}/${b}/something${c}fjfj/${dataSet}')
+        values << new ConfigurationValue(context.configuration, "complex", '/$USERHOME/$USERNAME/$USERGROUP')
```
vinjana: Should that not also contain exactly the part that was changed? I.e. such that values are not separated by '/'

```groovy
'/bli$USERHOME/bla$USERNAME/blub$USERGROUP'
```

and 

```groovy
'/${USERHOME}bli/${USERNAME}bla/${USERGROUP}blub'
```

vinjana: Also, what about incorrect variable replacement in
```groovy
'$USERNAMEbutDontReplaceThisUsernameButReportAsPlainDollarUsageWarning
```

RoddyCore/test/src/de/dkfz/roddy/config/validation/DefaultValidatorSpec.groovy
```diff
@@ -37,17 +52,46 @@ class DefaultValidatorSpec extends Specification {
...
-        'abc$'                  | 'string' | false    | [unattachedDollarCharacter.expand('bla')] | []
-        'abc${'                 | 'string' | false    | [unattachedDollarCharacter.expand('bla')] | []
-        'tr$ue'                 | 'string' | false    | [unattachedDollarCharacter.expand('bla')] | []
 
+        '\\$'                   | 'string' | true     | []                                      | []  //Escaped dollar
+        'abc\\$'                | 'string' | true     | []                                      | []  //Escaped dollar
+        'abc\\${'               | 'string' | true     | []                                      | []  //Escaped dollar with open brace
+        'a\\${b}c'              | 'string' | true     | []                                      | []  //Escaped dollar with complete braces
+
+        // Not ok
```
npeschke: If the ConfigurationIssue defines a detachedDollarCharacter with the following warning

> "Variable '#REPLACE_0#' contains plain dollar sign(s) without braces. Roddy does not interpret them as variables and cannot guarantee correct ordering of assignments for such variables in the job parameter file."

I would expect that a test for 
`'abc$'                  | 'string' | false    | [detachedDollarCharacter.expand('bla')] | []`
yields the same result as
`'$abc'                  | 'string' | false    | [detachedDollarCharacter.expand('bla')] | []`

however `$abc` throws no `detachedDollarCharacter` warning.

RoddyCore/test/src/de/dkfz/roddy/config/validation/DefaultValidatorSpec.groovy
```diff
@@ -37,17 +52,46 @@ class DefaultValidatorSpec extends Specification {
...
+        where:
+        value              | type     | expected | expectedWarnings                           | expectedErrors
+        'abc${var}'        | "string" | true     | []                                         | []
+        'abc\\${var}'      | "string" | true     | []                                         | []
+        'abc${var}v${var}' | "string" | true     | []                                         | []
+        '\\${c}'           | 'string' | true     | []                                         | []
+        '\\${c'            | 'string' | true     | []                                         | []
+        'abc${${var}}'     | "string" | false    | [inproperVariableExpression.expand('bla')] | []
+        'abc${'            | 'string' | false    | [inproperVariableExpression.expand('bla')] | []
+        'ab${c'            | 'string' | false    | [inproperVariableExpression.expand('bla')] | []
```
npeschke: If checks for opened braces are performed, why not test closed braces as well?
```
'abc${'            | 'string' | false    | [inproperVariableExpression.expand('bla')] | []
'abc$}'            | 'string' | false    | [inproperVariableExpression.expand('bla')] | []
'abc$def}'         | 'string' | false    | [inproperVariableExpression.expand('bla')] | []
```

RoddyCore/src/de/dkfz/roddy/config/ConfigurationIssue.groovy
```diff
@@ -37,15 +37,20 @@ class ConfigurationIssue {
      * Various templates for configuration related issues.
      */
     static enum ConfigurationIssueTemplate {
-        unattachedDollarCharacter(
+        detachedDollarCharacter(
                 ConfigurationIssueLevel.CVALUE_WARNING,
                 "Several variables in your configuration contain one or more dollar signs. As this might impose problems in your cluster jobs, check the entries in your job configuration files. See the extended logs for more information.",
-                "The variable named '#REPLACE_0#' contains one or more dollar signs, which do not belong to a Roddy variable definition (\${variable identifier}). This might impose problems, so make sure, that your results job configuration is created in the way you want."
+                "Variable '#REPLACE_0#' contains plain dollar sign(s) without braces. Roddy does not interpret them as variables and cannot guarantee correct ordering of assignments for such variables in the job parameter file."
```
vinjana: See comments on other PR. "contains unescaped dollar sign(s) without braces"

RoddyCore/src/de/dkfz/roddy/config/ConfigurationIssue.groovy
```diff
@@ -37,15 +37,20 @@ class ConfigurationIssue {
...
         ),
         valueAndTypeMismatch(
                 ConfigurationIssueLevel.CVALUE_ERROR,
                 "Several variables in your configuration mismatch regarding their type and value. See the extended logs for more information.",
-                "The value of variable named '#REPLACE_0#' does not match the variables type '#REPLACE_1#'."
+                "The value of variable named '#REPLACE_0#' is not of its declared type '#REPLACE_1#'."
+        ),
+        inproperVariableExpression(
+                ConfigurationIssueLevel.CVALUE_WARNING,
+                "Several variables in your configuration might misuse the variable import wrong. A variable needs to be imported into another value with \${variable identifier}.",
```
vinjana: dito

RoddyCore/src/de/dkfz/roddy/config/ConfigurationValue.groovy
vinjana: Sorry, I was unclear. From the code the evaluation order is clear. But it should be documented in the documentation (read the docs).

RoddyCore/src/de/dkfz/roddy/config/ConfigurationValue.groovy
```diff
@@ -466,12 +416,12 @@ class ConfigurationValue implements RecursiveOverridableMapContainer.Identifiabl
...
     @Deprecated
     boolean isInvalid() {
-        return invalid
+        return !valid
     }
 
     @Deprecated
     void setInvalid(boolean invalid) {
-//        this.invalid = invalid
+        // Not used anymore, invalid is not overriden.
```
vinjana: Hm. "isInvalid()" is still available (returns !valid) but setInvalid suddenly is a no-op. To me this sounds dangerous. Maybe better

```groovy
void setInvalid(boolean invalid) {
   valid = !invalid
}
```

RoddyCore/src/de/dkfz/roddy/config/loader/ConfigurationFactory.groovy
```diff
@@ -875,7 +878,7 @@ class ConfigurationFactory {
         //any directory to "path". In case of the output directories, this was a bad thing! So we know about
         //this specific type of variable and just set it to path at any time.
         if (key.endsWith("OutputDirectory"))
-            type = "path"
+            type = CVALUE_TYPE_PATH
```
vinjana: I was wondering. What is really meant by "path" is "directory". Maybe we should user `CVALUE_TYPE_DIRECTORY` and introduce a "directory" identifier into the XMLs and then in version 4 use "directory" in the XMLs?

vinjana: maybe "dirname", "directoryname" (in analogy to "filename"; I'd prefer camelcase though), "directoryName"

RoddyCore/src/de/dkfz/roddy/core/Analysis.groovy
vinjana: ?

RoddyCore/test/src/de/dkfz/roddy/config/ConfigurationValueSpec.groovy
vinjana: I think so, really. The whole point of the change is to make exactly this more generic. So a test would be appropriate.

RoddyCore/test/src/de/dkfz/roddy/config/validation/DefaultValidatorSpec.groovy
vinjana: At least now we have 'tr$ue' throws. So likely '$abc' throws either. To be 100% certain a dedicated '$true' could be added, but I would not expect much difference from 'tr$ue'.

RoddyCore/src/de/dkfz/roddy/config/loader/ConfigurationFactory.groovy
vinjana: I'll make a Roddy 4 issue out of this.

RoddyCore/test/src/de/dkfz/roddy/config/ConfigurationValueSpec.groovy
dankwart-de: The separator does not matter anymore, as the values are not getting split by the resolver. Also, these cases are already covered with value a.

dankwart-de: Configuration value issues tests are covered at another place!

RoddyCore/src/de/dkfz/roddy/config/ConfigurationValue.groovy
dankwart-de: Removed the method and the field

RoddyCore/src/de/dkfz/roddy/config/ConfigurationIssue.groovy
```diff
@@ -54,8 +54,8 @@ class ConfigurationIssue {
         ),
         inproperVariableExpression(
                 ConfigurationIssueLevel.CVALUE_WARNING,
-                "Several variables in your configuration might misuse the variable import wrong. A variable needs to be imported into another value with \${variable identifier}.",
-                "The variable named '#REPLACE_0#' might misuse the variable import. Make sure, that all intended variable imports are non-escaped and formed like \${variable identifier}. Also, variable nesting is not allowed."
+                "Several variables in your configuration might use malformatted variable imports. Variable imports must look like \${variable identifier}, nesting is forbidden.",
```
vinjana: 1. "Several". One is sufficient. Just "Variable(s)"
2. "might" expresses extreme uncertainty. Use "may" if you are really uncertain. Use "are" if you are certain.

RoddyCore/src/de/dkfz/roddy/config/ConfigurationIssue.groovy
```diff
@@ -54,8 +54,8 @@ class ConfigurationIssue {
...
+                "Variable '#REPLACE_0#' might use malformatted variable imports. Variable imports must look like \${variable identifier}, nesting is forbidden."
```
vinjana: dito

RoddyCore/src/de/dkfz/roddy/Constants.java
```diff
@@ -81,8 +75,11 @@
     public static final String PARAMETER_FILE_SUFFIX = ".parameters";
     public static final String PROJECT_NAME = "projectName";
     public static final String DATASET = "dataSet";
+    public static final String DATASET_HR = "dataset";
```
vinjana: What does HR mean?

RoddyCore/src/de/dkfz/roddy/core/Analysis.groovy
```diff
@@ -243,7 +235,7 @@ class Analysis {
             }
 
             if (!test && !canStartJobs(ds)) {
-                logger.postAlwaysInfo("The " + Constants.PID + " " + ds.getId() + " is still running and will be skipped for the process.")
+                logger.postAlwaysInfo("The ${Constants.DATASET_HR} ${ds.id} is still running and will be skipped for the process.")
```
vinjana: As I said, "dataset" in the conventional way is fine. "dataset" as identifier should be modelled as constant.

RoddyCore/test/src/de/dkfz/roddy/config/ConfigurationIssueSpec.groovy
```diff
@@ -41,7 +41,7 @@ class ConfigurationIssueSpec extends Specification {
 
         where:
         template                | messageContent | expectedLevel  | expectedCollectiveMessage
-        detachedDollarCharacter | smallArray     | CVALUE_WARNING | "Several variables in your configuration contain one or more dollar signs. As this might impose problems in your cluster jobs, check the entries in your job configuration files. See the extended logs for more information."
+        detachedDollarCharacter | smallArray     | CVALUE_WARNING | "Several variables contain plain dollar sign(s). Roddy does not interpret them as variables and cannot guarantee correct ordering of assignments for such variables in the job parameter file."
         valueAndTypeMismatch    | largeArray     | CVALUE_ERROR   | "Several variables in your configuration mismatch regarding their type and value. See the extended logs for more information."
```
vinjana: Concerning the extended log message. You are aware that we always have a reference to the extended log file for more information when there is any error. So this message here may be redundant.

Again: Keep the messages short and remove all junk words. "Several" here does not add any specific information. It is only important here __that__ there is/are variables.

docs/config/xmlConfigurationFiles.rst
```diff
@@ -128,17 +128,31 @@ The configuration value itself is defined as a cvalue element. Each element can
 * *type* - There exist several types for configuration values. The default value is string. Note, that the selection of
   the type will influence, how variables are interpreted and evaluated / converted.
 
-  - *string* will
+  - *string* accepts any value.
 
   - *int* will accept integer values only. E.g. 1, 2, 3 or 4.
 
-  - *float* will accept float values in differnt formats. E.g. 1.2f 1.2
+  - *float* will accept valid Java / Groovy float values. E.g. 1.2f 1.2
```
vinjana: I think the correct term here is "float literal", not "float value". 
I would remove Groovy. I think people will rather find the Java docs for literals, or not?

docs/config/xmlConfigurationFiles.rst
```diff
@@ -128,17 +128,31 @@ The configuration value itself is defined as a cvalue element. Each element can
...
 
+  - *boolean* will evaluate true (preferred), y, j, t and 1 to true and false (preferred), n, f and 0 to false.
 
+  - *path* indicates, that the variable is a path or a part of a path.
 
-  - *boolean*
+Nice to know - Order of evaluation in Roddy
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Configuration values can be cast to their *type* by calling any of the methods *toBoolean()*,
```
vinjana: Do the users of the XML know anything about the toFloat() etc. methods? Do they need to know?
Really focus on the declared type "float", "path", "file", etc. and explain the variable evaluation mechanism. Leave out Roddy interna.

RoddyCore/src/de/dkfz/roddy/SystemProperties.groovy
```diff
@@ -0,0 +1,38 @@
...
+ */
+package de.dkfz.roddy
+
+import groovy.transform.CompileStatic
+
+/**
+ * Facade for different system properties to reduce code duplication and typos.
+ */
+@CompileStatic
+class SystemProperties {
```
vinjana: Good idea to have that centralized!

vinjana:
BTW: Thanks for your patience! Thanks for the big refactoring and fix!

RoddyCore/src/de/dkfz/roddy/Roddy.java
```diff
@@ -773,7 +790,14 @@ public static void exit() {
 
     public static void exit(int ecode) {
         Initializable.destroyAll();
-        System.exit(ecode <= 250 ? ecode : 250); //Exit codes should be in range from 0 .. 255
+        if (jobManager != null)
+            jobManager.stopUpdateDaemon();
+        System.exit(ecode <= 255 ? ecode : ExitReasons.wrongExitCodeUsed.getCode()); //Exit codes should be in range from 0 .. 255
```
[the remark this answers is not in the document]
dankwart-de: #324

RoddyCore/src/de/dkfz/roddy/core/RuntimeService.groovy
```diff
@@ -460,18 +457,18 @@ class RuntimeService {
...
     }
 
     File getOutputAnalysisBaseDirectory(DataSet dataSet, Analysis analysis) {
-        return getOutputAnalysisBaseCV(analysis).toFile(analysis, dataSet)
+        return getOutputAnalysisBaseDirectoryCV(analysis).toFile(analysis, dataSet)
     }
 
     File getOutputAnalysisBaseDirectory(ExecutionContext context) {
-        return getOutputAnalysisBaseDirectory(context.dataSet, context.analysis)
+        return getOutputAnalysisBaseDirectoryCV(context.analysis).toFile(context)
```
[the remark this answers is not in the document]
dankwart-de: Yes, it does.

dankwart-de:
Thanks for your input! Better thorough than sorry! I'll fix the last things and merge everything then.

PR #314, opened 2018-11-26
Add roddy test spec class and change all copyright notices

dankwart-de:
Add the RoddyTestSpec class, update Copyright notices

Updated the copyright notice in all files to our current version.

This class RoddyTestSpec extends the Groovy Specification class and 
is used to do some basic initialization like the FileSystemProvier and 
the ExecutionService.

### DIFF
--- a/RoddyCore/src/com/stackoverflow/questions/xmlvalidation/LSInputImpl.java
+++ b/RoddyCore/src/com/stackoverflow/questions/xmlvalidation/LSInputImpl.java
@@ -1,7 +1,7 @@
 /*
  * Copyright (c) 2015 German Cancer Research Center (Deutsches Krebsforschungszentrum, DKFZ).
  *
- * Distributed under the MIT License (license terms are at https://www.github.com/eilslabs/Roddy/LICENSE.txt).
+ * Distributed under the MIT License (license terms are at https://www.github.com/TheRoddyWMS/Roddy/LICENSE.txt).
  */
 
 package com.stackoverflow.questions.xmlvalidation;
--- a/RoddyCore/src/com/stackoverflow/questions/xmlvalidation/LSInputImpl.java
+++ b/RoddyCore/src/com/stackoverflow/questions/xmlvalidation/LSInputImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 eilslabs.
+ * Copyright (c) 2015 German Cancer Research Center (Deutsches Krebsforschungszentrum, DKFZ).
  *
  * Distributed under the MIT License (license terms are at https://www.github.com/eilslabs/Roddy/LICENSE.txt).
  */
--- a/RoddyCore/src/de/dkfz/roddy/ExitReasons.groovy
+++ b/RoddyCore/src/de/dkfz/roddy/ExitReasons.groovy
@@ -11,7 +11,7 @@ import groovy.transform.CompileStatic
 @CompileStatic
 enum ExitReasons {
 
-    wrongStartupLocation(255, "You must call roddy from the right location! (Base roddy folder where roddy.sh resides."),
+    wrongStartupLocation(255, "Wrong execution directory. Change to directory with roddy.sh before execution."),
     groovyServError(254, "SystemExitException throw by GroovyServ, will exit now."),
     malformedCommandLine(253, "Command line was malformed."),
     unfulfilledRequirements(252, "Execution requirements unfulfilled."),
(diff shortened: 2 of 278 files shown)


RoddyCore/src/com/stackoverflow/questions/xmlvalidation/LSInputImpl.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 eilslabs.
+ * Copyright (c) 2015 German Cancer Research Center (Deutsches Krebsforschungszentrum, DKFZ).
  *
  * Distributed under the MIT License (license terms are at https://www.github.com/eilslabs/Roddy/LICENSE.txt).
```
vinjana: Wrong paths to Github repo (everywhere!)

https://www.github.com/TheRoddyWMS/Roddy/LICENSE.txt

RoddyCore/test/src/de/dkfz/roddy/RoddyTestSpec.groovy
```diff
@@ -0,0 +1,35 @@
...
+import org.junit.ClassRule
+import spock.lang.Specification
+
+/**
+ * Base class for Spock tests which does some basic initialization stuff.
+ */
+class RoddyTestSpec extends Specification {
+
+    @ClassRule
+    static final ContextResource contextResource = new ContextResource() {
```
vinjana: You said this resource would be available as a single instance and should not get modified -- in particular in parallelized tests, Did you check whether existing tests already modify this resource? Would it be possible to make this field non-final to allow to substitute the value by a copy? Furthermore this will force us to run parallel execution tests of tests (!). Can we not set the value in the subclass's specification and only provide the otherwise duplicated code here? E.g. make this here a trait to mix in with Specifications?

dankwart-de: I thought about it and I think it is better to just create a context when necessary. This way, we avoid issues when it comes to parallelization.

vinjana: What effect does this decision have on the code? This original code creating a ContextResource in the superclass is still in.

RoddyCore/src/de/dkfz/roddy/ExitReasons.groovy
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package de.dkfz.roddy
+
+import groovy.transform.CompileStatic
+
+@CompileStatic
+enum ExitReasons {
+
+    wrongStartupLocation(255, "You must call roddy from the right location! (Base roddy folder where roddy.sh resides."),
```
vinjana: Please try to make the messages as concise as possible.

"Wrong execution directory. Change to directory with roddy.sh before execution."

RoddyCore/src/de/dkfz/roddy/ExitReasons.groovy
```diff
@@ -0,0 +1,44 @@
...
+    malformedCommandLine(253, "Command line was malformed."),
+    unfulfilledRequirements(252, "Execution requirements unfulfilled."),
+    unparseableStartupOptions(251, "Startup options could not be parsed."),
+    unknownFeatureToggleFile(250, "Cannot find requested feature toggle file."),
+    unknownFeatureToggle(249, "Feature toggle is not known."),
+    unknownProxyProblem(248, "Unknown problem with proxy setup."),
+    scratchDirNotConfigured(247, Constants.APP_PROPERTY_SCRATCH_BASE_DIRECTORY + " is not defined."),
+    wrongJobManagerClass(246, "Wrong job manager class."),
+    appPropertiesFileNotFound(245, "Application properties file not found or loadable."),
+    analysisNotLoadable(244, "Could not load analysis"),
```
vinjana: terminal period missing

RoddyCore/src/de/dkfz/roddy/ExitReasons.groovy
```diff
@@ -0,0 +1,44 @@
...
+    severeConfigurationErrors(243, "Severe configuration errors occurred."),
+    unhandledException(242, "Unhandled exception."),
+    unknownSSHHost(241, "SSH remote host could not be reached."),
+    invalidSSHConfig(240, "SSH setup is not valid."),
+    fatalSSHError(239, "Fatal error during SSH setup."),
+    aJobHadAnError(238, "At least one job exited with an error."),
+    waitForJobsFailedWithAnUnknownError(237, "Roddy.waitForJobs() failed with an unknown exception."),
+
+    wrongExitCodeUsed(100, "Someone uses a wrong exit code somewhere in Roddy. Exit codes should be in class ExitReasons (if possible) and must be in the range [1;255].")
```
vinjana: Maybe just: "Exit codes should be in class ExitReasons (if possible) and must be in the range [1;255]."

I assume here, that we core devs do not any mistakes :-P

PR #315, opened 2018-11-26
Fix behaviour of syscalls to find in BashCommandSet

dankwart-de:
Change getListFilesInDirectoryCommand to use -L in the find method. This
will allow Roddy to follow symbolic (file) links. Directory links will
still be omitted.

Also apply this change to getListFullDirectoryContentRecursivelyCommand.
Change tests to match the new call to find.

Add tests for getListFilesInDirectoryCommand()

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/execution/io/fs/BashCommandSet.groovy
+++ b/RoddyCore/src/de/dkfz/roddy/execution/io/fs/BashCommandSet.groovy
@@ -186,7 +186,7 @@ class BashCommandSet extends ShellCommandSet {
 
     @Override
     String getListFilesInDirectoryCommand(File path) {
-        return "find ${path.absolutePath} -type f -maxdepth 1"
+        return "find -L ${path.absolutePath} -type f -maxdepth 1"
     }
 
     @Override
@@ -200,7 +200,7 @@ class BashCommandSet extends ShellCommandSet {
         StringBuilder sb = new StringBuilder("find ")
         sb << "\"" << f.absolutePath << "\""
 
-        if (depth > 0) sb << " -maxdepth ${depth}"
+        if (depth > 0) sb << " -L -maxdepth ${depth}"
         if (selectedType == FileType.DIRECTORIES) sb << " -type d"
         if (selectedType == FileType.FILES) sb << " -type f"
         if (detailed) sb << " -ls"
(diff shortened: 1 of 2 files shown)


RoddyCore/src/de/dkfz/roddy/execution/io/fs/BashCommandSet.groovy
```diff
@@ -200,7 +200,7 @@ class BashCommandSet extends ShellCommandSet {
         StringBuilder sb = new StringBuilder("find ")
         sb << "\"" << f.absolutePath << "\""
```
npeschke: To follow the specification of `find` it might be better to include the `-L` flag after find and before `f.absolutePath`

RoddyCore/src/de/dkfz/roddy/execution/io/fs/BashCommandSet.groovy
```diff
@@ -200,7 +200,7 @@ class BashCommandSet extends ShellCommandSet {
...
 
-        if (depth > 0) sb << " -maxdepth ${depth}"
+        if (depth > 0) sb << " -L -maxdepth ${depth}"
```
npeschke: Why set the `-L` flag only if depth > 0?

RoddyCore/test/src/de/dkfz/roddy/execution/io/fs/BashCommandSetSpecification.groovy
```diff
@@ -25,18 +25,36 @@ class BashCommandSetSpecification extends Specification {
...
 
         where:
         directory | depth | selectedType | complexList | result
-        tmpa      | -1    | ANY          | false       | 'find "/tmp/a"'
-        tmpa      | -1    | FILES        | false       | 'find "/tmp/a" -type f'
-        tmpa      | -1    | DIRECTORIES  | true        | 'find "/tmp/a" -type d -ls'
-        tmpa      | 0     | ANY          | false       | 'find "/tmp/a"'
-        tmpa      | 1     | FILES        | false       | 'find "/tmp/a" -maxdepth 1 -type f'
-        tmpa      | 2     | DIRECTORIES  | true        | 'find "/tmp/a" -maxdepth 2 -type d -ls'
+        tmpa      | -1    | ANY          | false       | 'find -L "/tmp/a"'
```
npeschke: The test fails for my build because the string is built differently in [Lines 200-208 of `BashCommandSet`](https://github.com/TheRoddyWMS/Roddy/pull/315/files#diff-749a2fe9bb0f19fdc8ef2823b54443bdR203)

- `-L` is appended after the path AND only if depth > 0

PR #316, opened 2018-11-26
Add a check for missing workflow classes

dankwart-de:
Add a check, if the workflow class of a loaded analysis exists. If the
class is not available, print an error message and abort.


dankwart-de:
Checks succeeded but did not report back. Merged it.

PR #325, opened 2018-12-10
Add first version of JSON configuration file loader and tests

dankwart-de:
Add the getResourceFile() method to the RoddyTestSpec class. This makes
import of resource files easy.

Add the test method testReadProcessingTools() to the
ConfigurationFactoryTest class. This is unfinished in this commit.

In ConfigurationFactory class:
- Add the loadJSONFile() and loadYAMLFile() method to the
- Add the selectProcessingtoolsNode() method to allow the lower case
  <processingtools> entry instead of <processingTools> in xml
  configuration files.

Add the JSONConverter class which can load a file and convert the string
to a valid Roddy xml configuration file. This way, we can facilitate our
existing xml checks and still use JSON and later YAML functionality.

Add the JSONConverterTest class and example configuration files. This
covers project and analsis xml configuration files. The test method will
run on a line by line test because a full text comparison did not work.

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/config/converters/JSONConverter.groovy
+++ b/RoddyCore/src/de/dkfz/roddy/config/converters/JSONConverter.groovy
@@ -1,0 +1,214 @@
+/*
+ * Copyright (c) 2017 German Cancer Research Center (Deutsches Krebsforschungszentrum, DKFZ).
+ *
+ * Distributed under the MIT License (license terms are at https://www.github.com/TheRoddyWMS/Roddy/LICENSE.txt).
+ */
+
+package de.dkfz.roddy.config.converters
+
+import de.dkfz.roddy.config.Configuration
+import de.dkfz.roddy.config.ConfigurationValue
+import de.dkfz.roddy.core.ExecutionContext
+import groovy.json.JsonSlurper
+import groovy.transform.CompileStatic
+import groovy.transform.TypeCheckingMode
+import groovy.xml.MarkupBuilder
+import sun.reflect.generics.reflectiveObjects.NotImplementedException
+
+/**
+ * In exception to our other Groovy classes, this class needs to be CompileDynamic!
+ * We use MarkupBuilder closures, where you cannot attach the CompileDynamic directive.
+ *
+ * Created by heinold on 30.01.17.
+ */
+class JSONConverter extends ConfigurationConverter {
+
+    @CompileStatic
+    @Override
+    String convert(ExecutionContext context, Configuration configuration) {
+        throw new NotImplementedException()
+    }
+
+    @CompileStatic
+    @Override
+    StringBuilder convertConfigurationValue(ConfigurationValue cv, ExecutionContext context) {
+        throw new NotImplementedException()
+    }
+
+    @CompileStatic
+    @Override
+    String convertToXML(File file) {
+        convertJsonString(file.text)
+    }
+
+    @CompileStatic
+    private String readAttribute(String attribute, def nc) { nc?.getAt(attribute) ?: "" }
+
+    @CompileStatic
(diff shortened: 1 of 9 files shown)


dankwart-de:
First version to discuss, more to come.
Who is integrating it into OTP? We should get their oppinion as well!

RoddyCore/src/de/dkfz/roddy/config/converters/JSONConverter.groovy
```diff
@@ -0,0 +1,214 @@
...
+
+    @CompileStatic
+    private String walltimeOf(def nc) { return readAttribute("walltime", nc) }
+
+    @CompileStatic(TypeCheckingMode.SKIP)
+    String convertJsonString(String json) {
+        /**
+         Remove comments first, they are not allowed in JSON and will produce errors.
+
+         See also a comment from Douglas Crockford:
```
vinjana: Really, this is just an opinion. This is the same argument like "Perl is bad, because you can write crappy hacks." I would not cite such an opinion in our code. The first comment line is sufficient and the solution to just ignore comments is fine -- although if you want to ignore comments (i.e. allow them in the JSON file), it is good to cite the source of he convention.

Actually, the inflation of braces and quotes and the lack of comments are exactly the reasons no to use JSON and prefer to the more powerful and expressive YAML (not that there are no arguments against YAML either).

RoddyCore/src/de/dkfz/roddy/config/converters/JSONConverter.groovy
```diff
@@ -0,0 +1,214 @@
...
+import de.dkfz.roddy.config.ConfigurationValue
+import de.dkfz.roddy.core.ExecutionContext
+import groovy.json.JsonSlurper
+import groovy.transform.CompileStatic
+import groovy.transform.TypeCheckingMode
+import groovy.xml.MarkupBuilder
+import sun.reflect.generics.reflectiveObjects.NotImplementedException
+
+/**
+ * In exception to our other Groovy classes, this class needs to be CompileDynamic!
```
vinjana: Not problematic, if the code is unit-tested.

RoddyCore/src/de/dkfz/roddy/config/converters/JSONConverter.groovy
```diff
@@ -0,0 +1,214 @@
...
+                                scriptparameter: scriptparameterOf(_output),
+                                filename: readAttribute("filename", _output)
+                        )
+                    }
+                }
+            }
+        }
+    }
+
+    Closure<MarkupBuilder> processResourcesets = { MarkupBuilder builder, def _tool ->
```
vinjana: Nice clear code!

dankwart-de:
Unfortunately, this is much more complicated than thought. The tests did not work at all and were in no way reliable. I had to figure out why and fixed them. Still needs some tweaks, but I am on a good way. I'll tell you when I think its reviewable again. I'll consider your comments though!

RoddyCore/src/de/dkfz/roddy/config/converters/JSONConverter.groovy
dankwart-de: This statement is from the Json inventor. I don't like that he did it but I understand why. That's why I cited it. But I follow his idea and put in comments which will be removed. Actually I only need them for copyright headers.

dankwart-de:
Ok, I pushed again. You can take a fresh look at it.

vinjana:
To be tested:
* value starting with brace open `{` -- these may be interpreted by a parser as dict-beginning. Also test the case `{...}`, but meant as string, not as dict.
* special characters and escaping
  * `\\`
  * `\$`
  * `\"`

PR #329, opened 2019-01-14
Fix regression related not-quoting bash/map variables provided as arg…

vinjana:
…uments to Job-constructor.

In-line parameters of job-calls are maps and were not correctly mapped to quoted string in parameter file resulting in not exporting them into the Bash scripts (due to bash bug).

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/config/ConfigurationValue.groovy
+++ b/RoddyCore/src/de/dkfz/roddy/config/ConfigurationValue.groovy
@@ -404,7 +404,7 @@ class ConfigurationValue implements RecursiveOverridableMapContainer.Identifiabl
     }
 
     private List<String> _bashArrayToStringList() {
-        String vTemp = value.trim().substring(1, value.length() - 2).trim() //Split away () and leading or trailing white characters.
+        String vTemp = value.trim()[1 .. -3].trim() //Split away () and leading or trailing white characters.
         String[] temp = vTemp.split(SPLIT_WHITESPACE) //Split by white character
 
         //Ignore leading and trailing brackets
--- a/RoddyCore/src/de/dkfz/roddy/config/ConfigurationValue.groovy
+++ b/RoddyCore/src/de/dkfz/roddy/config/ConfigurationValue.groovy
@@ -404,7 +404,7 @@ class ConfigurationValue implements RecursiveOverridableMapContainer.Identifiabl
     }
 
     private List<String> _bashArrayToStringList() {
-        String vTemp = value.substring(1, value.length() - 2).trim() //Split away () and leading or trailing white characters.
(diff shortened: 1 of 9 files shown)


RoddyCore/src/de/dkfz/roddy/config/ConfigurationValue.groovy
```diff
@@ -404,7 +404,7 @@ class ConfigurationValue implements RecursiveOverridableMapContainer.Identifiabl
     }
 
     private List<String> _bashArrayToStringList() {
-        String vTemp = value.substring(1, value.length() - 2).trim() //Split away () and leading or trailing white characters.
+        String vTemp = value.trim().substring(1, value.length() - 2).trim() //Split away () and leading or trailing white characters.
```
dankwart-de: value.trim()[1 .. -3].trim()

RoddyCore/src/de/dkfz/roddy/config/converters/BashConverter.groovy
```diff
@@ -245,12 +244,20 @@ class BashConverter extends ConfigurationConverter {
             return '"' + value + '"'
     }
 
-    static String convertListToBashArrayString(List list) {
-        "(${list.collect { it.toString() }.join(" ")})" as String
+    static String convertListToBashArrayString(List list, doQuote = true) {
+        String result = "(${list.collect { it.toString() }.join(" ")})"
+        if (doQuote)
```
dankwart-de:
``` lang=java
if (doQuote)
  return "\"${result}\""  ?
return result
```

RoddyCore/src/de/dkfz/roddy/config/converters/BashConverter.groovy
```diff
@@ -245,12 +244,20 @@ class BashConverter extends ConfigurationConverter {
...
+            '"' + result + '"'
+        else
+            result
     }
 
-    static String convertMapToBashMapString(Map map) {
-        "(${map.collect { k, v -> "[$k]=$v"}.join(" ")})" as String
+    static String convertMapToBashMapString(Map map, doQuote = true) {
+        String result = "(${map.collect { k, v -> "[$k]=$v" }.join(" ")})"
+        if (doQuote)
```
dankwart-de: same here, two returns and no else ?

RoddyCore/src/de/dkfz/roddy/config/converters/BashConverter.groovy
vinjana: Think this is only a minor gain over 

```groovy
if (cond)
  "a"
else
  "b"
```

One could also argue that from a Groovy perspective the current code is more idiomatic, as it does not use unnecessary `return` statements.

PR #340, opened 2019-03-21
Fix a bug with BaseFile.getFilename()

dankwart-de:
Bump to 3.5.2

In BaseFile
- In the BaseFile constructor, add another if, which checks for the
  availability of multiple parent files. The code looks complicated
  but directly accessing _helper.parentFiles lead to a Cast exception.
  This effectively makes the setParentFiles call obsolete in some cases.
- Make setParentFiles without selectionTag deprecated
- Add the selectionTag parameter to setParentFiles

In GenericMethod
- Remove the call to setParentFiles, this functionality is now available
  in the constructor and uses the CreationHelper instance.


RoddyCore/src/de/dkfz/roddy/knowledge/files/BaseFile.groovy
```diff
@@ -345,7 +345,11 @@ abstract class BaseFile<FS extends FileStageSettings> extends FileObject {
 
         if (helper instanceof ConstructionHelperForGenericCreation) { // Manual creation is currently intrinsic.
             ConstructionHelperForGenericCreation _helper = helper as ConstructionHelperForGenericCreation
-            if (_helper.parentObject instanceof FileGroup) {
+            def _pFiles = helper.parentFiles
```
vinjana: `_pFiles` -> `parents`

This "_" notation is hardly ever used -- even in our code. pFiles is also much less understandible.

RoddyCore/src/de/dkfz/roddy/knowledge/files/BaseFile.groovy
```diff
@@ -549,16 +553,17 @@ abstract class BaseFile<FS extends FileStageSettings> extends FileObject {
...
     }
 
+    @Deprecated
     void setParentFiles(List<BaseFile> parentfiles) {
-        setParentFiles(parentfiles, false)
+        setParentFiles(parentfiles, false, DEFAULT)
     }
 
-    void setParentFiles(List<BaseFile> parentfiles, boolean resetFilename) {
+    void setParentFiles(List<BaseFile> parentfiles, boolean resetFilename, String selectionTag) {
```
vinjana: I am wondering whether handing over `selectionTag` from method to method should be classified as bad smell and an indication of a design issue ...
Should it maybe be a field of BaseFile?

RoddyCore/src/de/dkfz/roddy/knowledge/methods/GenericMethod.groovy
```diff
@@ -554,8 +554,9 @@ class GenericMethod {
         if (!fileParameter.checkFile.evaluate(context))
             bf.setAsTemporaryFile()
 
-        if (allInputFiles.size() > 1)
-            bf.setParentFiles(allInputFiles, true)
+        // Should already be done with the helper.
```
vinjana: Why keep the comment?

dankwart-de: As a reminder for you, the reviewer to look over it.

RoddyCore/src/de/dkfz/roddy/knowledge/files/BaseFile.groovy
dankwart-de: As far as I remember, in some cases, the tag is overriden by the workflow developer, that is why you can pass it.

RoddyCore/src/de/dkfz/roddy/knowledge/files/BaseFile.groovy
vinjana: Removed the temporary variable. Only used twice.

RoddyCore/src/de/dkfz/roddy/knowledge/methods/GenericMethod.groovy
vinjana: Done in the constructor using the "ConstructionHelper" parameter object:

de.dkfz.roddy.knowledge.files.BaseFile#BaseFile(de.dkfz.roddy.knowledge.files.BaseFile.ConstructionHelperForBaseFiles)

Note that there `this.path` is set to the saved old value, if the filename-pattern resolution fails. In this constructor/new instance context a resetting to on old value is equivalent to setting to the default value and is not necessary.

vinjana:
Hi @dankwart-de. How did this Basefile bug appear to the users? For the changelogs ...

PR #351, opened 2020-10-07
Fix LocalExecutionService ignored errors

vinjana:
LocalExecutionService ignored errors when running asynchronously with waitFor=false. This may have been responsible for non-detection of errors during bresume calls and accumulation of PSUPS LSF-cluster-jobs.

This fix uses the AsyncExecutionResult to wrap asynchronously computed exit code, stdout, & stderr. Currently, errors are only reported via log-messages if ignoreErrors = true but produce an UnexpectedExecutionResultException if ignoreErrors = false.

Additional smaller changes are commented in the PR.


RoddyCore/src/de/dkfz/roddy/execution/io/ExecutionService.groovy
```diff
@@ -101,7 +106,7 @@ abstract class ExecutionService implements BEExecutionService {
         }
     }
 
-    ExecutionService() {
+    protected ExecutionService() {
```
vinjana, line 109: The constructor is not used except in the initialize method above. This is a singleton and the proper way to initialize it seems to call the initialize method, not the constructor directly.

RoddyCore/src/de/dkfz/roddy/execution/io/LocalExecutionService.groovy
```diff
@@ -10,16 +10,30 @@ import de.dkfz.roddy.SystemProperties
...
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
+import java.util.concurrent.Future
+import java.util.function.Consumer
+import java.util.function.Supplier
 
 /**
  * The local execution service executes commands on the local machine. For this groovy's execute() method is used.
- * @author michael
+ * All commands are executed using Bash as interpreter!
```
vinjana, line 27: This was not the case before. The change corrects the inconsistent behaviour.

RoddyCore/src/de/dkfz/roddy/execution/io/LocalExecutionService.groovy
```diff
@@ -10,16 +10,30 @@ import de.dkfz.roddy.SystemProperties
...
  */
-@groovy.transform.CompileStatic
+@CompileStatic
 class LocalExecutionService extends ExecutionService {
+
     private static final LoggerWrapper logger = LoggerWrapper.getLogger(LocalExecutionService.class.name)
 
+    // Note that the thread pool is unbounded.
+    private ExecutorService executorService = Executors.newCachedThreadPool()
```
vinjana, line 35: Note that the cache is unbounded.

RoddyCore/src/de/dkfz/roddy/execution/io/fs/FileSystemAccessProvider.groovy
```diff
@@ -630,7 +630,7 @@ class FileSystemAccessProvider {
     boolean setAccessRightsRecursively(File path, String accessStringDirectories, String accessString, String group) {
         return ExecutionService.instance.
                 execute(commandSet.getSetAccessRightsRecursivelyCommand(
-                        path, accessStringDirectories, accessString, group), false)
+                        path, accessStringDirectories, accessString, group))
```
vinjana: This will continue to run asynchronously, but with better error reporting.

RoddyCore/src/de/dkfz/roddy/execution/jobs/Job.groovy
```diff
@@ -172,7 +172,7 @@ class Job extends BEJob<BEJob, BEJobResult> {
                 , inlineScript ? null : getToolMD5(TOOLID_WRAPIN_SCRIPT, context)
                 , getResourceSetFromConfiguration(toolID, context)
                 , []
-                , [:]
+                , [:] as Map<String,String>
```
vinjana, line 175: Never Groovy (2.4.19) didn't like the old way.

build.gradle
```diff
@@ -92,7 +92,7 @@ repositories {
 }
 
 
-ext.roddyToolLibVersion = "2.1.1"
+ext.roddyToolLibVersion = "2.2.0"
```
vinjana, line 95: See https://github.com/TheRoddyWMS/RoddyToolLib/pull/27

RoddyCore/src/de/dkfz/roddy/plugins/LibrariesFactory.groovy
```diff
@@ -56,7 +56,7 @@ class LibrariesFactory extends Initializable {
 
     private Map<PluginInfo, File> loadedJarsByPlugin = [:]
 
-    private PluginInfoMap mapOfPlugins = [:];
+    private PluginInfoMap mapOfPlugins = new PluginInfoMap([:])
```
vinjana, line 59: A bug gone unnoticed.

juleskers (commented):
Comments from yesterday, round 1

Still checking the rest together with @jpalbrecht

RoddyCore/test/src/de/dkfz/roddy/execution/io/LocalExecutionServiceSpec.groovy
```diff
@@ -0,0 +1,72 @@
...
+        ExecutionResult res = es.execute("echo 'hello'; false", false)
+
+        then:
+        res.exitCode == 1
+        !res.successful
+        res.resultLines.size() == 1
+        res.resultLines[0] == "hello"  // Note that the newline is removed.
+    }
+
+    def "execute('sleep 5; false', waitFor = false)"() {
```
juleskers, line 52: the name is a lie! :smiley: (5 != 2)

RoddyCore/test/src/de/dkfz/roddy/execution/io/LocalExecutionServiceSpec.groovy
```diff
@@ -0,0 +1,72 @@
...
+    @Rule
+    final ContextResource contextResource = new ContextResource()
+
+    static {
+        ExecutionService.initializeService(LocalExecutionService, RunMode.CLI)
+    }
+
+    LocalExecutionService es = ExecutionService.instance
+
+    def "execute('echo')"() {
```
juleskers, line 19: As discussed in meeting: these names can be better :smile:

Roddy.iml
```diff
@@ -1,81 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
vinjana: Actually, I did not notice and I am very surprised, because apparently IntelliJ does not need this. The file is gone, but the project module -- now simply imported from the Gradle model -- is present. Not sure yet, what that might mean.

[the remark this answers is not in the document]
vinjana: I am more or less the only developer left and the system needed to be changed, because IntelliJ apparently does not support the old way with one module including sources and tests. All this is now automatically modelled as submodules and there is no way to change this, because IntelliJ automatically reimports from the Gradle-file.

I suggest we leave this like it is here in the PR.

RoddyCore/src/de/dkfz/roddy/execution/io/LocalExecutionService.groovy
vinjana: What is your opinion?

RoddyCore/src/de/dkfz/roddy/execution/io/LocalExecutionService.groovy
```diff
@@ -38,25 +51,72 @@ class LocalExecutionService extends ExecutionService {
...
+                                       boolean ignoreErrors,
+                                       OutputStream outputStream = null) {
         if (waitFor) {
             return LocalExecutionHelper.executeCommandWithExtendedResult(command, outputStream)
         } else {
-            Thread.start {
-                command.execute()
-            }
-            return new ExecutionResult(true, 0, [], "")
+            ProcessBuilder processBuilder = new ProcessBuilder(["bash", "-c", command])
```
vinjana, line 85: The old implementation used `String.execute(String)`, which eventually refers to `Runtime.exec(String command, String[] end, File dir)`. This latter method tokenizes the command using the delimiters space, tab, newline, carriage-return, form-feed. Bash, by contrast, uses the delimiters space, tab, newline by default (the `$IFS` variable), which, however, in this context here boils down to the same, because here only spaces are relevant token-delimiters. This means that in our context, `command.execute()` and `bash -c ...` called via `ProcessBuilder` will do the same.

RoddyCore/src/de/dkfz/roddy/execution/io/LocalExecutionService.groovy
vinjana: See the comparison of `String.execute()` and `bash -c` in a `ProcessBuilder`, below.

RoddyCore/src/de/dkfz/roddy/execution/io/fs/BashCommandSet.groovy
```diff
@@ -122,33 +127,79 @@ class BashCommandSet extends ShellCommandSet {
...
+        if (null != fileGroup)
+            result += "chgrp ${fileGroup} ${path}"
+        return Optional.ofNullable(result == "" ? null : result)
     }
 
     @Override
-    String getSetAccessRightsRecursivelyCommand(File f, String rightsForDirectories, String rightsForFiles, String fileGroup) {
-        def path = "${f.getAbsolutePath()}"
-        return "find ${path} -type d | xargs chmod ${rightsForDirectories}; find ${path} -type d | xargs chgrp ${fileGroup}; find ${path} -type f | xargs chmod ${rightsForFiles}; find ${path} -type f | xargs chgrp ${fileGroup};"
+    Optional<String> getSetAccessRightsRecursivelyCommand(File f, String rightsForDirectories, String rightsForFiles, String fileGroup) {
```
vinjana, line 185: This here (and below) fixes the problem with `null` parameter values provided to these methods in the old Roddy code and failing calls like `chgrp null $file` that resulted from it.

RoddyCore/src/de/dkfz/roddy/execution/io/fs/FileSystemAccessProvider.groovy
```diff
@@ -622,15 +622,21 @@ class FileSystemAccessProvider {
...
+     * @param accessStringDirectories    permission string for directories (string, such as "rwx", or octal)
+     * @param accessStringFiles          permission string for files (string, such as "rwx", or octal)
+     * @param group                      group name
      * @return
      */
-    boolean setAccessRightsRecursively(File path, String accessStringDirectories, String accessString, String group) {
-        return ExecutionService.instance.
-                execute(commandSet.getSetAccessRightsRecursivelyCommand(
-                        path, accessStringDirectories, accessString, group), false)
+    boolean setAccessRightsRecursively(File path, String accessStringDirectories, String accessStringFiles, String group) {
```
vinjana, line 636: Adaptation to deal with Optional return value.

RoddyCore/src/de/dkfz/roddy/execution/io/fs/FileSystemAccessProvider.groovy
```diff
@@ -639,12 +645,25 @@ class FileSystemAccessProvider {
...
+     * @return
+     */
+    boolean setAccessRights(File file, String accessString, String group) {
         ExecutionService eService = ExecutionService.instance
         if (eService.canModifyAccessRights()) {
-            return eService.modifyAccessRights(file, accessString, groupID)
+            return eService.modifyAccessRights(file, accessString, group)
         } else {
-            return eService.execute(commandSet.getSetAccessRightsCommand(file, accessString, groupID))
+            return commandSet.getSetAccessRightsCommand(file, accessString, group)
```
vinjana, line 664: dito

RoddyCore/test/src/de/dkfz/roddy/execution/io/LocalExecutionServiceSpec.groovy
```diff
@@ -16,52 +16,55 @@ class LocalExecutionServiceSpec extends Specification {
...
         res.exitCode == 1
         !res.successful
-        res.resultLines.size() == 1
+        res.resultLines.size() == 2
         res.resultLines[0] == "hello"  // Note that the newline is removed.
+        res.resultLines[1] == "error"
     }
 
-    def "execute('echo \"hello\"; false', waitFor = false)"() {
+    def "execute synchronously and succeed with empty stdout and non!-ignored stderr"() {
```
jpalbrecht, line 45: shouldn't that be a "execute synchronously and *fail*" (exit code = 1)?

RoddyCore/test/src/de/dkfz/roddy/execution/io/LocalExecutionServiceSpec.groovy
```diff
@@ -16,52 +16,55 @@ class LocalExecutionServiceSpec extends Specification {
 
     LocalExecutionService es = ExecutionService.instance
 
-    def "execute('echo')"() {
+    def "execute synchronously and succeed with captured stdout and non!-ignored stderr"() {
```
juleskers, line 19: JPA and I checked all the names, thanks for improving them!
One seems to have run into a copy-paste error (see comment below), but otherwise good :+1:

juleskers:
Question: There is a seemingly empty merge-commit in the history.
Is that an accidental artifact?

OK either way if you keep or remove.

RoddyCore/src/de/dkfz/roddy/execution/io/fs/BashCommandSet.groovy
```diff
@@ -122,20 +127,47 @@ class BashCommandSet extends ShellCommandSet {
...
+     *  the group ownership can be changed to the target group.
+     *
+     * @param file    path to check. Not exactly this path is changed, but a file besides that path.
+     * @param group   target-group to check; if this is null, then only chmod is tested, otherwise also chgrp.
+     * @return
+     */
+    @Override
+    String getCheckChangeOfPermissionsPossibilityCommand(File file, String group) {
+        File testFile = new File(file, ".roddyPermissionsTestFile")
+        if (null == group) {
```
jpalbrecht, line 166: is Yoda conditions default in Roddy? Seen this at various positions.

juleskers (approved):
Pair-reviewed together with @jpalbrecht ;

Good to merge, excluding a few minor remarks, and one incorrect spec-name (copy paste error);
All the other things are basically stylistic things that are also fine as-is.

RoddyCore/src/de/dkfz/roddy/execution/io/ExecutionService.groovy
```diff
@@ -769,7 +769,9 @@ abstract class ExecutionService implements BEExecutionService {
                             // Unzip the file again. foundExisting stays true
                             GString str = RoddyIOHelperMethods.getCompressor().getDecompressionString(remoteZipFile, analysisToolsServerDir, analysisToolsServerDir)
                             getInstance().execute(str, true)
-                            provider.setDefaultAccessRightsRecursively(new File(analysisToolsServerDir.getAbsolutePath()), context)
+                            boolean success = provider.setDefaultAccessRightsRecursively(new File(analysisToolsServerDir.getAbsolutePath()), context)
+                            if (!success)
+                                logger.warning("Command error occurred and was ignored")
```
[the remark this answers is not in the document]
vinjana: Too true to be true.

RoddyCore/src/de/dkfz/roddy/execution/io/ExecutionService.groovy
```diff
@@ -792,7 +794,9 @@ abstract class ExecutionService implements BEExecutionService {
 
                     GString str = RoddyIOHelperMethods.getCompressor().getDecompressionString(new File(dstCommonExecutionDirectory, remoteFile.getName()), analysisToolsServerDir, analysisToolsServerDir)
                     getInstance().execute(str, true)
-                    provider.setDefaultAccessRightsRecursively(new File(analysisToolsServerDir.getAbsolutePath()), context)
+                    boolean success = provider.setDefaultAccessRightsRecursively(new File(analysisToolsServerDir.getAbsolutePath()), context)
+                    if (!success)
+                        logger.warning("Command error occurred and was ignored")
```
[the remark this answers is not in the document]
vinjana: made it dependent on the if/else condition

RoddyCore/src/de/dkfz/roddy/execution/io/LocalExecutionService.groovy
vinjana: Not sure what the thumb-up means. "Good discussion"? "Leave it like it is"? "Poor guy. Keep up the good mood!" ;-)

RoddyCore/src/de/dkfz/roddy/execution/io/fs/BashCommandSet.groovy
vinjana: I try to use these "Yoda conditions" (thanks for the term) to not by chance write `group = null`. Note that `group == null` evaluates to `true` if group is null, but `group = null` evaluates to `false`. So, in case of a typo with `=` rather than `==`, not only `group` would be assigned, but also the other branch would be taken. Just a safety measure.

RoddyCore/test/src/de/dkfz/roddy/execution/io/LocalExecutionServiceSpec.groovy
vinjana: Sorry. Maybe a tabular test would have been more appropriate for all these parameter combinations and results.

RoddyCore/src/de/dkfz/roddy/execution/io/fs/BashCommandSet.groovy
vinjana: Actually, I never checked, but apparently Groovy does not accept assignments in conditions. So it is not strictly unnecessary. However, the "added cognitive load" counter-argument on the Wiki page is ridiculously irrelevant in the context of the overall complexity of the code. So I leave it as it is but avoid it in the future. Thanks for pointing out.

PR #355, opened 2021-04-27
Fix bresume 1 error

vinjana:
- Better error reporting for submission errors
  - Bugfix: Failure to pass job resuming step (e.g. via `bresume`), with jobs that cannot be resumed (e.g. in the Bam2FastqPlugin). Try resume only with jobs in HOLD or SUSPENDED states.
  - The command in `roddyCall.sh` is now properly escaped and suited for direct copy-paste to the command-line
  - Stack-traces are now only reported to the `.tsv` log in the `.roddy/` directory and a message is reported to provide this file to the maintainers for bug-reports
  - Bumped used BatchEuphoria to 0.0.13
  - Bumped used RoddyToolLib to 2.3.0
  - This will be Roddy 3.6.1

Some more explanations:

  - Removed some try-catch blocks, to ensure exceptions are handled further up the call stack (and reported!) and execution is really halted in case of fatal problems.
  - Many endless lines were broken into more readable code
  - Deprecated bloaty Job-constructors. Probably a builder would be the optimal solution for constructing these objects, but this seemed overkill for software in pure maintenance mode.

Like the BE PR this one will only succeed in the CI, if RoddyToolLib and BE have been released.


GWarsow (approved):
Only minor remarks or quesrtions.

vinjana (commented):
I noticed that if I start Roddy now in a different situation, somehow jobs are not released. This is unfortunate. I'll have to do some debugging first :-(

PR #356, opened 2021-05-04
Removed runtimeConfig.sh and runtimeConfig.xml from execution store.

vinjana:


vinjana:
Are you fine with removing the `runtimeConfig.sh` and `runtimeConfig.xml`?

GWarsow:
> Are you fine with removing the `runtimeConfig.sh` and `runtimeConfig.xml`?

Fine for me if it causes confusion otherwise.

GWarsow (approved):
> Are you fine with removing the `runtimeConfig.sh` and `runtimeConfig.xml`?

Fine for me if it causes confusion otherwise.

PR #357, opened 2021-07-30
Changes for API-breaking BE update. Reimplementation of some ExecutionServices

vinjana:
* patch: Removed redundant implementation of LocalExecutionService (now calling RoddyToolLib code)
* patch: Added timeout arguments to ExecutionService.execute method calls, defaulting to Duration.ZERO (= no timeout)
* patch: SSHExecutionService now uses CompletableFutures for execution thus allowing for asynchronous execution.
* patch: JobState.STARTED considered as kind of running.
* patch: removed dysfunctional `ignoreErrors` from Roddy's ExecutionService implementation. Error are not ignored thus, but by catching and handling exceptions.

Note that the implementation of the SSHExecutionService timeout is a workaround using the Linux command `timeout`. The reason is that there is a version conflict for SSHJ caused by the JSCH agent-proxy command. SSHJ and JSCH libraries could eventually be replaced by something else, e.g. mina-ssh.

### DIFF
--- a/RoddyCore/src/de/dkfz/roddy/config/validation/DefaultValidator.groovy
+++ b/RoddyCore/src/de/dkfz/roddy/config/validation/DefaultValidator.groovy
@@ -100,7 +100,7 @@ class DefaultValidator extends ConfigurationValueValidator {
         return result
     }
 
-    static final String VARIABLE = '([^\\\\{][$][{]([\\w[-]_]){1,}[}])|(^[$][{]([\\w[-]_]){1,}[}])'
+    static final String VARIABLE = '(^|(?<=[^\\\\{]))[$][{]([\\w[-]_]){1,}[}]'
     static final String VARIABLE_START = '([^\\\\|^{][$][{])|(^[$][{])'
 
     boolean areVariablesProperlyDefined(ConfigurationValue cv) {
@@ -115,7 +115,7 @@ class DefaultValidator extends ConfigurationValueValidator {
         boolean result = countOfVariables == countOfStartedVariables
 
         if (!result) {
-            super.warnings << new ConfigurationIssue(inproperVariableExpression, cv.id, cv?.configuration?.file?.toString())
+            super.warnings << new ConfigurationIssue(malformedVariableExpression, cv.id, cv?.configuration?.file?.toString())
         }
 
         return result
--- a/RoddyCore/src/de/dkfz/roddy/execution/io/SSHExecutionService.groovy
+++ b/RoddyCore/src/de/dkfz/roddy/execution/io/SSHExecutionService.groovy
@@ -395,9 +395,6 @@ class SSHExecutionService extends RemoteExecutionService {
         CompletableFuture<Tuple3<Integer,List<String>, List<String>>> processF = CompletableFuture.supplyAsync({
             SSHPoolConnectionSet connectionSet = waitForService()
             SSHClient sshClient = connectionSet.client
-
-            // Append a newgrp (Philip: better "newgrp -"!) to each command, so that all
-            // command context in the proper group context.
             connectionSet.acquire()
             Session session = sshClient.startSession()
             List<String> stdout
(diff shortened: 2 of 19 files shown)


RoddyCore/src/de/dkfz/roddy/execution/io/SSHExecutionService.groovy
```diff
@@ -353,83 +361,84 @@ class SSHExecutionService extends RemoteExecutionService {
...
-
+            List<String> stdout
+            List<String> stderr
+            Integer exitCode
             try {
-                cmd = session.exec(command)
+                Session.Command executingCommand
+                if (timeout != Duration.ZERO) {
+                    // TODO Remove this workaround. Try switching to mina-sshd.
+                    Long seconds = Math.round(Math.ceil(timeout.toNanos() / 1000000000.0d))
```
vinjana, line 410: This is rounding up to full seconds. `timeout` cannot go lower than seconds.

RoddyCore/src/de/dkfz/roddy/execution/io/SSHExecutionService.groovy
```diff
@@ -353,83 +361,85 @@ class SSHExecutionService extends RemoteExecutionService {
...
+                    Long killSeconds = 60
```
vinjana, line 411: The `timeout` command first sends a SIGTERM, such that the job may clean up things, but after 60 seconds SIGKILL, to get the job done in reasonable time.

RoddyCore/src/de/dkfz/roddy/execution/io/SSHExecutionService.groovy
```diff
@@ -353,83 +361,85 @@ class SSHExecutionService extends RemoteExecutionService {
...
+                    String timeoutCommand = "timeout --preserve-status -s TERM -k ${killSeconds}s ${seconds}s"
+                    executingCommand = session.exec("${timeoutCommand} ${command}")
+                } else {
+                    executingCommand = session.exec(command)
+                }
+                if (outputStream) {
+                    stdout = LocalExecutionHelper.readStringStream(executingCommand.inputStream, outputStream)
```
vinjana, line 418: Note that this implementation differs from the one in RoddyToolLib. The main reason is that I didn't know the details about output streaming with SSHJ. I therefore decided to basically stick to old (working) implementation with synchronous readout of output streams.

vinjana:
Note that Travis is not used for CI anymore (relatively limited resources are provided, which turned out to be limiting). I didn't manage to remove it from this list of required tests yet.

RoddyCore/src/de/dkfz/roddy/config/validation/DefaultValidator.groovy
```diff
@@ -1,123 +1,123 @@
...
+        boolean result = !temp.findAll(matchDetachedDollars)
+
+        if (!result) {
+            super.warnings << new ConfigurationIssue(detachedDollarCharacter, cv.id, cv?.configuration?.file?.toString())
+        }
+
+        return result
+    }
+
+    static final String VARIABLE = '(^|(?<=[^\\\\{]))[$][{]([\\w[-]_]){1,}[}]'
```
vinjana, line 103: This fixes a small bug when matching variable references in configuration values. See the tests in:

https://github.com/TheRoddyWMS/Roddy/pull/357/files?diff=split&w=1#diff-3b522784d7050e976297f3f71b1591edc1c0cd14c4ef5c28f11ff92bc72fb32c

RoddyCore/src/de/dkfz/roddy/execution/io/SSHExecutionService.groovy
```diff
@@ -353,83 +361,85 @@ class SSHExecutionService extends RemoteExecutionService {
...
+                             Duration timeout = Duration.ZERO,
+                             OutputStream outputStream = null,
+                             ExecutorService executorService) {
+
+        CompletableFuture<Tuple3<Integer,List<String>, List<String>>> processF = CompletableFuture.supplyAsync({
+            SSHPoolConnectionSet connectionSet = waitForService()
+            SSHClient sshClient = connectionSet.client
+
+            // Append a newgrp (Philip: better "newgrp -"!) to each command, so that all
+            // command context in the proper group context.
```
askask, line 400: This comment doesn't make sense to me, where is `newgrp` added (and shouldn't it be prepended)?

Otherwise your changes look good

vinjana: I think the comment is not from me and you're right in that it is misplaced and unclear. I will just remove it. I remember the discussion, though. The aim was to run commands such that files are already created with the right group. We never followed up on this.

vinjana:
If you are satisfied with the changes, please resolve the discussions and approve the MR. You may also just merge, if you wish.

askask (approved):

PR #359, opened 2022-01-20
Updated proxy-vole to use a newer fork.

vinjana:
Originally, in the hope to solve some license problem with Eclipse Xtend Runtime Library 2.8.3 in its dependencies. The Fossa verification error however could be resolved by explicitly entering the multilicense of proxy-vole (MIT|Apache 2.0|GLPv3). An update to a newer versions may still be a good idea.


vinjana:
I will lift the changes to the current code and include them into another PR.

